PR #1, opened 2020-05-21
[build][docs] Add package.json and readme

niedzielski:
Add boilerplate for NPM and readme.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1,0 +1,30 @@
+# 🧩 @wikimedia/mw-vue-components
+
+Vue.js user interface component library prototype for MediaWiki's Vector skin.
+
+## Table of contents
+
+<!--
+	Markdown Preview Enhanced is used to automatically generate the table of contents. You don't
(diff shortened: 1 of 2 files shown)


readme.md
```diff
@@ -0,0 +1,30 @@
+# 🧩 @wikimedia/mw-vue-components
+
+Vue.js user interface component library prototype for MediaWiki's Vector skin.
```
hknustwmf, line 3: Maybe we should incorporate the Vector constraint into the repo name?

Volker-E: Aren't we planning to have these components as general Vue.js UI components for MediaWiki mid-term?

niedzielski: Migrated discussion to [T253357](https://phabricator.wikimedia.org/T253357) and added renaming as an acceptance criterion. Naming shouldn't block development IMHO but is expected to conclude Tuesday-ish.

PR #7, opened 2020-06-11
Add jest configuration and a simple test. Updated readme.md

speedandfunction-anribolon:


.gitignore
```diff
@@ -0,0 +1,6 @@
+.idea
```
niedzielski, line 1: Let's keep this file as precise as possible. So far as I can see, the only file needed for this change is `/node_modules/` and maybe `/coverage/`? I do think we should add excellent documentation where useful to the readme on configuring one's editor (VS Code, vim, ...) but would prefer to avoid making IDE-specific changes to the code.

Regarding OS-specific files, these should be configured at a system level by the Git user, in my view. If an OS litters the filesystem with temporaries, like "Thumbs.db", these can be excluded in a global Git excludesFile per user, not the project-specific .gitignore, as there is no repo where these temporary files would actually be wanted in version control.

https://git-scm.com/docs/gitignore

.jest/jest.config.js
```diff
@@ -0,0 +1,20 @@
+module.exports = {
```
niedzielski, line 1: Can this file be plain JSON or are we using some functional code here?

package.json
```diff
@@ -17,7 +17,27 @@
...
+    "vue": "^2.6.11"
+  },
+  "devDependencies": {
+    "@babel/core": "7.10.2",
+    "@types/jest": "26.0.0",
+    "@vue/test-utils": "1.0.3",
+    "babel-core": "6.26.3",
+    "babel-preset-vue": "2.0.2",
+    "jest": "26.0.0",
+    "ts-jest": "^26.0.0",
```
niedzielski, line 35: Would you mind pinning to the latest package patch version here so we can easily track versions without looking at package-lock? So, `"26.1.0"`.

package.json
```diff
@@ -17,7 +17,27 @@
   "scripts": {
     "preversion": "[ -z \"$(git status -z)\" ]",
     "version": "npm t",
+    "test:unit": "jest -c .jest/jest.config.js",
+    "test": "npm -s run test:unit",
```
niedzielski, line 21: If possible, please consider moving the more important scripts to the top. In this case, `test` much more important than `preversion` or `version` scripts. I should have moved `preversion` below `version`.

.jest/jest.setup.js
```diff
@@ -0,0 +1,16 @@
+/* eslint-disable no-implicit-globals */
```
niedzielski, line 1: Would you mind removing this file for now? I like that you've thought about this and especially appreciate the documentation. However, I think we should discuss this further when it is actually needed. If our production code requires global dependencies injected like this, it probably means that those dependencies are implicit which is bad. I would like to make it a little hard on us to add a global implicit dependency and this file makes it easy.

.jest/jest.config.js
```diff
@@ -0,0 +1,20 @@
...
+  transform: {
+    '.+\\.ts$': 'ts-jest',
+    '.+\\.vue$': 'vue-jest',
+    '^.+\\.(c|le)ss$': '<rootDir>/.jest/ignoreTransform.js',
+  },
+  'snapshotResolver': '<rootDir>/.jest/snapshotResolver.js',
+  'setupFiles': [
+    '<rootDir>/.jest/jest.setup.js'
+  ],
+  collectCoverage: true,
```
niedzielski, line 12: Is it easy to generate an HTML report is that more work?

.jest/snapshotResolver.js
```diff
@@ -0,0 +1,11 @@
+module.exports = {
+    resolveSnapshotPath( testPath, snapshotExtension ) {
+        return testPath.replace( /\.test\.([tj]sx?)/, `${snapshotExtension}.$1` );
```
niedzielski, line 3: I don't think we are likely to have JSX or TSX files in this repo. Please remove the `x`.

package.json
```diff
@@ -17,7 +17,27 @@
...
     "postversion": "npm publish",
     "prepublishOnly": "git push origin \"$(git rev-parse --abbrev-ref @)\" \"$(git tag --points-at @)\""
+  },
+  "peerDependencies": {
+    "vue": "^2.6.11"
+  },
+  "devDependencies": {
```
niedzielski, line 28: Wow! Are all of these dependencies needed for this Jest configuration? I was surprised to see babel and ts-loader.

readme.md
```diff
@@ -73,6 +75,36 @@ npm install
 ```
 </details>
 
+### Testing
+
+#### Unit Tests
+- WVUI uses [Vue Test Utils](https://vue-test-utils.vuejs.org/), the official unit testing utility library for Vue.js.
```
niedzielski, line 81: I don't think we have any markdown linters configured but please consider limiting line length to 100 characters.

package.json
speedandfunction-anribolon: you're right, we don't need both

.jest/jest.config.js
speedandfunction-anribolon: @niedzielski It actually generates an html report. It creates `coverage` dir and generated a report there. Not sure if we need to commit `coverage` dir, what do you think?

niedzielski: I think this would be valuable to publish as documentation so it's easy to compare local results to results on master. Maybe we can publish to docs/coverage? I don't think we should commit it but CI can build it for us on commit when we have that configured.

speedandfunction-anribolon: Sounds good for me!

.jest/jest.config.js
speedandfunction-anribolon: I believe we MAY use some code here in the future. I would keep it as a js file. Is there any reason to replace it with JSON?

.gitignore
speedandfunction-anribolon: I agree

.gitignore
```diff
@@ -0,0 +1,2 @@
+coverage
```
niedzielski, line 1: It's my understanding from your comment below that this contains an HTML report. Can this be under docs/coverage to indicate that it should be served? I tried to do this even for source maps on the prototype ([example](https://mw-components.netlify.app/sourcemaps/mwcmingzip)).

Also, would you mind anchoring these to the root as directories? Like:

```
/docs/coverage/
/node_modules/
```

.jest/jest.config.js
niedzielski: I think JSON is more appropriate because it clarifies that file is a simple configuration that can be read by any JSON parser. JavaScript is a program and nontrivial to parse. I would prefer to generally assume the simplest possible implementations unless a need for complexity is imminent.

.jest/jest.config.js
```diff
@@ -0,0 +1,20 @@
+module.exports = {
+  rootDir: '..',
+  transform: {
+    '.+\\.ts$': 'ts-jest',
+    '.+\\.vue$': 'vue-jest',
+    '^.+\\.(c|le)ss$': '<rootDir>/.jest/ignoreTransform.js',
```
niedzielski, line 6: Sorry, I missed this earlier but I don't think the `^` is adding anything.

types/ViewLoader.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module '*.vue' {
```
niedzielski, line 1: A minor typo in the filename. Please consider calling this V**ue**Loader.d.ts.

src/components/button/Button.vue
```diff
@@ -0,0 +1,19 @@
+<template>
```
niedzielski, line 1: Looks good to me. We'll need to add a spot to put a label pretty soon. Maybe:

```
    <button @click="onClick">
        <slot />
    </button>
```

This can be done later though.

readme.md
```diff
@@ -73,6 +75,38 @@ npm install
...
+|-- src
+    |-- components
+        |-- your-component
+            |-- YourComponent.vue
+            |-- YourComponent.test.ts
+            |-- YourComponent.snap.ts
+```
+- Coverage report will be created automatically in `coverage` directory.
+
+To run tests, use `npm run test` command (see [NPM scripts](#npm-scripts)).
```
niedzielski, line 107: Although this works I think `npm test` is more conventional.

package.json
niedzielski: Do we need babel-core and babel-preset-vue?

.jest/snapshotResolver.js
```diff
@@ -0,0 +1,11 @@
+module.exports = {
+    resolveSnapshotPath( testPath, snapshotExtension ) {
+        return testPath.replace( /\.test\.([tj]s?)/, `${snapshotExtension}.$1` );
+    },
+
+    resolveTestPath( snapshotFilePath, snapshotExtension ) {
```
niedzielski, line 6: Consider renaming from `snapshotFilePath` to `snapshotPath`. This matches what's used in the `resolveSnapshotPath()` method above and the naming of both methods themselves.

.jest/jest.setup.js
niedzielski: Any thoughts on this, @speedandfunction-anribolon?

package.json
niedzielski: I think this question is still open.

.gitignore
niedzielski: Thoughts on rooting with a leading slash and indicating directory with a trailing slash?

PR #14, opened 2020-06-15
[docs] add initial Less naming conventions

niedzielski:
Add BEM for the initial Less naming conventions. The convention may
change dramatically based on the outcome of T255100 but is at least the
interim convention to use until told otherwise.

The link is HTTP because there is no HTTPS site.

Bug: T255100


speedandfunction-anribolon:
Looks good to me, I believe using BEM is a good idea.

PR #34, opened 2020-06-29
[dev] Format JSON and Markdown better

niedzielski:
Wikimediaify the JSON and Markdown files.

Bug: T256693


changelog.md
```diff
@@ -4,34 +4,35 @@ Versions and bullets are arranged chronologically from latest to oldest.
 
 ## v0.0.1 (unreleased)
```
Volker-E, line 6: Without looking at this with a Markdown viewer (not sure if GitHub diffing is showing me everything), doesn't this change the rendering?  
In OOUI  we leave them unintended

.prettierrc.json
```diff
@@ -0,0 +1,6 @@
+{
```
Volker-E, line 1: Out of curiosity. Wouldn't it be better to use an eslint plugin for this?

niedzielski: In my experience, Prettier's ability to apply automatic formatting is far superior to ESLint (all formatting changes in this patch were applied automatically without exception, for example). It also supports many languages without custom community maintained plugins of varying quality (eslint-plugin-json is used in this repo, for example, but it supports very few options). Finally, Prettier is ESLint compatible.

If you haven't tried it out yet, please do! [The Prettier playground](https://prettier.io/playground/) is a great place to start. I've been using it for a couple years and haven't seen anything quite like it. In a nutshell, Prettier is unbelievably good at accepting utter garbage code in and generating beautiful and extremely consistent outputs. Its changed the way I read and write code for repositories where its used.

changelog.md
niedzielski: This is tab width of four. I see no difference locally or in GitHub's rich diff.

diff

.prettierrc.json
speedandfunction-anribolon: In addition, Prettier is compatible with the most of [editors/IDEs](https://prettier.io/docs/en/editors.html)

Volker-E: @speedandfunction-anri Interesting, which IDEs are incompatible with eslint?

PR #41, opened 2020-07-05
[update] Flesh out Button props

niedzielski:
Add some of the component properties and tests needed by Button.

Just the structure. Needs some linter changes coming soon. Styles in new PR.

Bug: T256038


src/components/button/Button.vue
```diff
@@ -1,16 +1,44 @@
 <template>
-	<button class="wvui-button" @click="onClick">
+	<button :class="classes" @click="onClick">
```
niedzielski, line 2: I will address whatever class approach we want to take in the styles patch.

src/components/button/PrimaryAction.ts
```diff
@@ -0,0 +1,32 @@
+/**
```
AnneTee, line 1: @niedzielski Should this live in the button component? It could also be used by Icon and any other components that use these common flags.

niedzielski: @AnneTee, fixed :+1:

AnneTee: Thanks! ☀️

src/components/button/Button.stories.ts
```diff
@@ -1,17 +1,83 @@
...
+					{ enabled: true, quiet: false },
+					{ enabled: true, quiet: true },
+					{ enabled: false, quiet: false },
+					{ enabled: false, quiet: true }
+				]
+			};
+		},
+		methods: { click: action( 'click' ) },
+		template: `
+		<table>
```
Volker-E, line 56: Minor comment: We should aim for no `table` in the presentation. `dl` would be an alternative. Can follow later…

src/actions/PrimaryAction.ts
```diff
@@ -0,0 +1,32 @@
...
+   */
+  Default = 'default',
+  /**
+   * The consequence of this action is to proceed to the next step in or conclude the current
+   * process. For example, creation of a page or confirmation that .
+   */
+  Progressive = 'progressive',
+  /**
+   * The consequence of this action is irreversible, data loss, or is difficult to undo. For
+   * example, deleting a page, discarding a draft edit, or blocking a user. **Never** use
```
Volker-E, line 18: Is this `**Never**` markdown? What's the output channels for them?

niedzielski: We use the Markdown JSDoc plugin in the Popups repo which I think (but am unsure) allows Markdown anywhere in a JSDoc. Regardless, just like rendered Markdown, the intent was to emphasize.

src/components/button/Button.stories.ts
niedzielski: I went the table route just to avoid introducing story-only styles and the data was tabular. I don't feel strongly about it.

PR #42, opened 2020-07-05
[component][button] Add initial styles

niedzielski:
The following seven rules appear to break lengthy rules, such as multi-property
transitions and mixins, that must appear across multiple lines:

- declaration-colon-space-after
- declaration-block-semicolon-newline-before
- declaration-block-semicolon-space-before
- value-list-comma-newline-after
- function-parentheses-space-inside
- function-comma-newline-after
- function-parentheses-newline-inside

What to do?

The presentation seems pretty close to the filled styles in the OOUI
button showcase.

Is the class name and cascade structure wanted?

Can we keep theming isolated in wikimedia.less for now and sort out full
theming later?

Bug: T256038


.stylelintrc.json
```diff
@@ -2,6 +2,13 @@
 	"extends": "stylelint-config-wikimedia",
 	"plugins": ["stylelint-no-unsupported-browser-features"],
 	"rules": {
-		"plugin/no-unsupported-browser-features": "error"
```
niedzielski, line 5: @Volker-E, do you know a better way? I'm hitting issues with lengthy rules like `transtition`. My preference is to _not_ use a `// stylelint-disable-line max-len` on these but that means writing them differently or updating the configuration. I've taken the latter approach but it's not ideal.

src/components/button/Button.vue
```diff
@@ -47,8 +47,184 @@ export default Vue.extend( {
 
 <style lang="less">
 @import ( reference ) '~wikimedia-ui-base/wikimedia-ui-base.less';
+@import ( reference ) '@/themes/wikimedia.less';
```
Volker-E, line 50: Interesting. Like to see this. 
But the theme should probably be called WikimediaUI. And reference itself the base.

src/components/button/Button.vue
```diff
@@ -47,8 +47,184 @@ export default Vue.extend( {
...
+		box-shadow @transition-base
+	;
+
+	&::-moz-focus-inner {
+		// [Firefox] hide the focus ring in all variations.
+		border: 0;
+	}
+
+	&:focus {
+		// Hide the standard focus outline.
```
Volker-E, line 79: Maybe that should say, “…outline. But enable it accessible further down.”

src/components/button/Button.vue
```diff
@@ -47,8 +47,184 @@ export default Vue.extend( {
...
+		outline: 0; // stylelint-disable-line plugin/no-unsupported-browser-features
+	}
+
+	&[ disabled ] {
+		// Hide the border but keep the sizing.
+		border: @border-width-base @border-style-base #0000;
```
Volker-E, line 85: Will file a task. No hex RGBA colors, browser support isn't there. `transparent` is your cross-browser supported friend.

changelog.md
```diff
@@ -4,6 +4,7 @@ Versions and bullets are arranged chronologically from latest to oldest.
 
 ## v0.0.1 (unreleased)
 
+-   [update] Add initial Button styles
```
Volker-E, line 7: In OOUI we've prepended the commit with the component affected in order to make orientation while skimming and searching simple:
“[update] Button: Add inital styles”

src/components/button/Button.vue
```diff
@@ -47,8 +47,184 @@ export default Vue.extend( {
 
 <style lang="less">
 @import ( reference ) '~wikimedia-ui-base/wikimedia-ui-base.less';
+@import ( reference ) '@/themes/wikimedia.less';
 
 .wvui-button {
-	background: @wmui-color-base20;
+	box-sizing: border-box; // All components use border-box box-sizing.
```
Volker-E, line 53: Captured in https://phabricator.wikimedia.org/T257239

niedzielski: https://github.com/wikimedia/wvui/pull/43

src/components/button/Button.vue
niedzielski: Actually, I learned these are compiled to `transparent` by cssnano. I think we should keep these as hex in these source files anticipation of support but it's a not sticking point.

src/components/button/Button.vue
niedzielski: I've taken some liberties with your suggestion. Happy to revise.

changelog.md
```diff
@@ -4,8 +4,8 @@ Versions and bullets are arranged chronologically from latest to oldest.
 
 ## v0.0.1 (unreleased)
 
--   [update] Add initial Button styles
--   [update] Flesh out Button props
+-   [update][button] Add initial styles
```
Volker-E, line 7: Is it a tag?

src/components/button/Button.vue
```diff
@@ -46,11 +46,11 @@ export default Vue.extend( {
 </script>
 
 <style lang="less">
-@import ( reference ) '~wikimedia-ui-base/wikimedia-ui-base.less';
-@import ( reference ) '@/themes/wikimedia.less';
+@import ( reference ) '@/themes/wikimedia-ui.less';
```
Volker-E, line 49: Nice!

changelog.md
niedzielski: Let's discuss on https://github.com/wikimedia/wvui/pull/42#discussion_r450403322

src/components/button/Button.vue
```diff
@@ -46,11 +46,11 @@ export default Vue.extend( {
...
 
 .wvui-button {
-	box-sizing: border-box; // All components use border-box box-sizing.
+	// All components use border-box box-sizing.
```
Volker-E, line 52: Why is a note about all components in the button component? I think it's fine to add `box-sizing` as is, w/o the comment.

src/components/button/Button.vue
```diff
@@ -46,11 +46,11 @@ export default Vue.extend( {
...
+	box-sizing: border-box;
 	// Interactive elements have a minimum and maximum touch area.
```
Volker-E, line 54: The “maximum” area isn't there for touch, it's for dealing with UI weirdness like having a button taking up a full canvas width. Not sure if the maximum comment is needed, it's definitely confusing

src/components/button/Button.vue
niedzielski: Good point. Fixed.

src/components/button/Button.vue
niedzielski: I've just dropped the max-width comment.

src/components/button/Button.vue
```diff
@@ -76,7 +79,7 @@ export default Vue.extend( {
 	}
 
 	&:focus {
-		// Hide the standard focus outline.
+		// Hide the standard focus outline. A border and box-shadow representation is added below.
```
Volker-E, line 80: Great addition.

src/themes/wikimedia-ui.less
```diff
@@ -1,3 +1,5 @@
+@import (reference) '~wikimedia-ui-base/wikimedia-ui-base.less';
```
Volker-E, line 1: Add whitespace in brackets. Also what does the `~` stand for?

.stylelintrc.json
Volker-E: An example from OOUI:
```
		.oo-ui-transition(
			height @transition-dialog,
			width @transition-dialog,
			top @transition-dialog,
			left @transition-dialog,
			right @transition-dialog,
			bottom @transition-dialog
		);
```
Perfect for me as it's a preprocessor mixin, not real CSS, where whitespace breaks in values makes me very unhappy.

niedzielski: Note that we won't be using mixins as much with the vendor prefixing. As I understand it, we're talking about these two options (unless you know others):

Option 1:

```less
	transition: border-color @transition-base, background-color @transition-base, color @transition-base, box-shadow @transition-base; // stylelint-disable-line max-len
```

Option 2:

```less
	transition:
		border-color @transition-base,
		background-color @transition-base,
		color @transition-base,
		box-shadow @transition-base
	;
```

You prefer option 1?

Volker-E: I don't think all rules necessary to be overriden. Don't want to end up in 
```
font-family: 
  Arial,
  Helvetica,
  'Helvetica Neue',
  sans-serif
;
```
land. Rather add the transition property values on one line. Alternatively we could consider making an exception for transitions. Transition is the exception and shouldn't be the rule.

src/themes/wikimedia-ui.less
niedzielski: This means the path is relative `node_modules`. I've added some notes to the readme.

niedzielski: (Whitespace should be fixed now too.)

src/components/button/Button.vue
Volker-E: Fascinating, why so? :) I've always found hex colors hard to grasp, the same is true for RGBA hex colors. `transparent` is a common English word and a unique identifier.

niedzielski:
I think this is ready for merging. @AnneTee, you may want to merge master into your icon patch once this PR is gets in.

changelog.md
Volker-E: Every patch is an update. For someone only slightly familiar “update” either means nothing or anything. Would [component] fulfill your need better? Just a thought.

niedzielski: Sure, I can get behind component.

src/components/button/Button.vue
niedzielski: Sorry, lost this conversation. I've replaced #0000 with transparent.

.stylelintrc.json
niedzielski: This is "resolved" in the sense that I've added what I think is the lesser evil of the two workarounds, option 1 without a lint disable:

```
	transition: border-color @transition-base, background-color @transition-base, color @transition-base, box-shadow @transition-base;
```

A better solution is wanted (and welcome) in the future.

Volker-E:
@niedzielski Still think that a pattern “[component] Button: Add initial styles” or “[component] Input: blah” etc might help to quickly skim through an update to the library to look for the components applied in library consumer's interface.

niedzielski:
@Volker-E, please continue discussion in https://github.com/wikimedia/wvui/pull/42#pullrequestreview-443311564. I've already revised to `[component][Button]`. Please squash and merge this patch if it's ready to go.

PR #43, opened 2020-07-07
[build][dev] Enable CSS automatic vendor prefixing

niedzielski:
- Add PostCSS Loader. A dummy `transition` is added to Button ensure it
  gets prefixed.

  I'm unsure if this will conflict with
  stylelint-no-unsupported-browser-features. If we find out it does, we
  can try to configure it or uninstall
  stylelint-no-unsupported-browser-features. I didn't see any docs on an
  stylelint-no-unsupported-browser-features with autoprefix
  configuration.

  https://github.com/postcss/autoprefixer/blob/5acf784/README.md#options
  https://github.com/postcss/postcss-loader

- Update the docs!

The CSSNano + PostCSS + Webpack medley is complex. This may require more
tweaking later but appears to be working correctly for our simple setup.

Bug: T257239


webpack.config.js
```diff
@@ -100,25 +108,6 @@ const config = ( _env, argv ) => ( {
 	// this source code is publicly distributed.
 	devtool: argv.mode === 'production' ? 'source-map' : 'cheap-module-eval-source-map',
 
-	optimization: {
```
niedzielski, line 103: In case it's unclear, this is now using the default Webpack optimization settings and the new postcss-loader above.

bundlesize.config.json
```diff
@@ -1,6 +1,6 @@
 {
 	"files": [
 		{ "path": "./dist/wvui.js", "maxSize": "1.1 KB" },
-		{ "path": "./dist/wvui.css", "maxSize": "0.1 KB" }
+		{ "path": "./dist/wvui.css", "maxSize": "0.2 KB" }
```
niedzielski, line 4: For convenience, here's the output generated:

```css
.wvui-button{background:#54595d;color:transparent;-webkit-transition:all 1s;transition:all 1s}
/*# sourceMappingURL=wvui.css.map */
```

PR #46, opened 2020-07-10
[update] Add initial TextInput styles and props.

speedandfunction-anribolon:
* Add basic <input/> component
* Most of native HTML element attributes such as `id`, `title`, `name` are bound to input element, except `disabled` and `type`, which are used to define some css classes for styling
* Add styles from OOUI
* Add tests for TextInput and TextInputType


src/components/text-input/TextInputType.ts
```diff
@@ -0,0 +1,17 @@
+/*
+* Defines types for text inputs
```
niedzielski, line 2: We want great docs so please consider linking to the MDN documentation here and also noting that this subset of supported types controls the underlying `input` element.

https://developer.mozilla.org/docs/Web/HTML/Element/Input#<input>_types

src/themes/wikimedia-ui.less
```diff
@@ -12,7 +12,19 @@
...
 
 @border-radius-base: 3px;
 
 @line-height-base: unit(18 / @font-size-browser / @font-size-base, @wvui-unit);
+
+@border-color-input--hover: @border-color-base--active;
+
+@box-shadow-widget: inset 0 0 0 1px transparent;
+
+.wvui-placeholder( @rules ) {
```
niedzielski, line 26: Why do we need this mixin? How is `.wvui-placeholder({ color: red; });` better than just writing `&::placeholder { color: red; }`?

If we keep it:
1. Do you think this should be part of the theme or is it a more general mixin?
2. Please add documentation for its intended usage.

src/components/text-input/TextInput.vue
```diff
@@ -0,0 +1,130 @@
...
+			text-shadow: @text-shadow-base--disabled;
+			border-color: @border-color-base--disabled;
+		}
+
+		.wvui-placeholder({
+			color: @color-placeholder;
+			opacity: 1;
+		});
+
+		&::-ms-clear {
```
niedzielski, line 105: I see some notes on MDN about this. Please consider adding some documentation about what this is used for and, if applicable, when it can be deleted.

https://developer.mozilla.org/docs/Web/CSS/::-ms-clear

src/components/text-input/TextInput.test.ts
```diff
@@ -0,0 +1,21 @@
+import { shallowMount } from '@vue/test-utils';
+import WvuiTextIput from './TextInput.vue';
+
+describe( 'wvui-text-input', () => {
```
niedzielski, line 4: Is this outer describe useful? I ended up dropping the outer describe for the Button tests. Everything in the file is concerned with wvui-button which was already being grouped by Jest.

https://github.com/wikimedia/wvui/blob/aa08dff4fa5a3de1a585a05319b28b19d2adf69e/src/components/button/Button.test.ts

package-lock.json
```diff
@@ -21306,7 +21306,8 @@
           "version": "2.1.1",
```
niedzielski, line 21306: Should this be committed? We didn't change anything.

src/components/text-input/TextInput.snap.ts
```diff
@@ -0,0 +1,14 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`wvui-text-input matches the snapshot 1`] = `
+<div
+  class="wvui-text-input"
+>
+  <input
+    aria-autocomplete="list"
+    autocomplete="off"
+    class="wvui-text-input-field wvui-text-input-field--enabled wvui-text-input-field--search"
```
niedzielski, line 10: Do we need `wvui-text-input-field--enabled` and `wvui-text-input-field--search`? Why can't we use the `input` attributes like `&:not( [ disabled ] )` and `&[ type='search' ] `?

src/components/text-input/TextInput.stories.ts
```diff
@@ -0,0 +1,37 @@
...
+			change: action( 'change' )
+		},
+		template: `
+		<wvui-text-input
+			:placeholder="placeholder"
+			:type="type"
+			:disabled="disabled"
+			@input="input"
+			@change="change"
+		></wvui-text-input>
```
niedzielski, line 35: I think we can use a self-closing tag `<wvui-text-input ... />`.

src/components/text-input/TextInput.vue
```diff
@@ -0,0 +1,130 @@
...
+		}
+	}
+} );
+</script>
+
+<style lang="less">
+@import ( reference ) '@/themes/wikimedia-ui.less';
+
+.wvui-text-input {
+	position: relative;
```
niedzielski, line 65: Would you mind adding a comment for why relative positioning is needed?

src/components/text-input/TextInput.stories.ts
Volker-E: @niedzielski As we're compiling components, are we going away from the MediaWiki syntax in some/many/every aspect/s? Regarding the self-closing syntax discussion and change recently in RL by Roan…

src/components/text-input/TextInput.vue
```diff
@@ -0,0 +1,130 @@
...
+			@change="onChange"
+		>
+	</div>
+</template>
+
+<script lang="ts">
+import { TextInputType, isTextInputType } from './TextInputType';
+import Vue, { PropType } from 'vue';
+
+export default Vue.extend( {
```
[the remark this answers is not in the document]
Volker-E: Was wondering something similar? 
What other inputs are we going to expect? 

search-input?
password-input?
tel-input?
email-input?
textarea?

Would like to be sure if we're doing ourselves a favor making it text-input…

src/components/text-input/TextInput.vue
```diff
@@ -0,0 +1,130 @@
...
+				'wvui-text-input': true
+			};
+		},
+		classes(): Record<string, boolean> {
+			return {
+				'wvui-text-input-field': true,
+				'wvui-text-input-field--disabled': this.disabled,
+				'wvui-text-input-field--enabled': !this.disabled,
+				'wvui-text-input-field--search': this.type === TextInputType.search,
+				'wvui-text-input-field--text': this.type === TextInputType.text
```
Volker-E, line 46: Here the separation of text-input and search-input already gets more difficult. For simplicity,  I'd rather refer to `input type=checkbox` as `wvui-checkbox` over `wvui-checkbox-input`

src/components/text-input/TextInput.vue
```diff
@@ -0,0 +1,130 @@
...
+
+<style lang="less">
+@import ( reference ) '@/themes/wikimedia-ui.less';
+
+.wvui-text-input {
+	position: relative;
+	vertical-align: middle;
+	box-sizing: border-box;
+
+	.wvui-text-input-field {
```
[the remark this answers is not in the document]
Volker-E: That should be one of the beauties of BEM to circumvent these descendant  selectors. 
You should be able to keep CSS hierarchy flat and lean thanks to it.
`.wvui-text-input-field` as most upper selector in contrast to `.wvui-text-input .wvui-text-input-field` or .wvui-text-input > .wvui-text-input-field`.
The only time where this could turn into an issue is when other components are brought in into a slot, an input inheriting an embedded input.

src/themes/wikimedia-ui.less
speedandfunction-anribolon: You're right it does not make much sense as we have autoprefixes

.stylelintrc.json
```diff
@@ -2,6 +2,18 @@
 	"extends": "stylelint-config-wikimedia",
 	"plugins": ["stylelint-no-unsupported-browser-features"],
 	"rules": {
-		"plugin/no-unsupported-browser-features": "error"
+		"plugin/no-unsupported-browser-features": [
```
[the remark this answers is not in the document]
speedandfunction-anribolon: @niedzielski Correct me if I'm wrong but I believe you're confusing this rule with max-len.

src/components/text-input/TextInput.vue
```diff
@@ -0,0 +1,130 @@
+<template>
+	<div :class="wrapperClasses">
```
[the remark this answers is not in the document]
speedandfunction-anribolon: Per Anne's [suggestion](https://github.com/wikimedia/wvui/pull/47#discussion_r453007157) let's keep classes "static" for now instead of using computed properties.

src/components/text-input/TextInput.test.ts
speedandfunction-anribolon: Good idea.

src/components/text-input/TextInput.snap.ts
speedandfunction-anribolon: I was thinking about some edge cases, but nothing came to my mind, actually. Removed

src/components/text-input/TextInput.vue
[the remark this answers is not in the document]
speedandfunction-anribolon: I've renamed it to wvui-input.

src/components/text-input/TextInput.vue
```diff
@@ -0,0 +1,130 @@
...
+	},
+	computed: {
+		wrapperClasses(): Record<string, boolean> {
+			return {
+				'wvui-text-input': true
+			};
+		},
+		classes(): Record<string, boolean> {
+			return {
+				'wvui-text-input-field': true,
```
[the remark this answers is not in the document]
speedandfunction-anribolon: Thanks for pointing this out!

src/components/text-input/TextInput.vue
```diff
@@ -0,0 +1,130 @@
+<template>
+	<div :class="wrapperClasses">
+		<input
+			autocomplete="off"
```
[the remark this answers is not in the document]
speedandfunction-anribolon: Yeah you're right. I've removed hardcoded attribute.

.stylelintrc.json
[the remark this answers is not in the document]
niedzielski: `max-len` and these rules are related, I thought. At least the only reason I can recall adding these rules is be cause transition specifically was exceeding the maximum line length. If not, then we don't need them?

src/components/text-input/TextInput.stories.ts
niedzielski: Generally speaking, I don't think we should limit ourselves to the capabilities of on-the-fly compilation. Otherwise, we may as well not pre-compile.

src/components/input/Input.snap.ts
```diff
@@ -0,0 +1,27 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`matches the snapshot Case 0 Input Type: text : ({"type": "text"}) => HTML 1`] = `
+<div
+  class="wvui-input"
+>
+  <input
+    aria-autocomplete="list"
```
niedzielski, line 8: Do we want `aria-autocomplete` still? I don't think it applies to all inputs.

src/components/input/InputType.test.ts
```diff
@@ -0,0 +1,19 @@
+import { InputType, isInputType } from './InputType';
+
+describe( 'isTextInputType()', () => {
```
niedzielski, line 3: I think these should be revised for the new name, `isInputType()`.

src/entries/wvui.ts
```diff
@@ -1,5 +1,7 @@
 import WvuiButton from '@/components/button/Button.vue';
+import WvuiTextInput from '@/components/input/Input.vue';
```
niedzielski, line 2: New name (here and elsewhere).

src/components/input/InputType.ts
```diff
@@ -0,0 +1,17 @@
...
+* Defines types for text inputs
+* */
+export enum InputType {
+	text = 'text',
+	search = 'search'
+}
+
+/**
+ * @param val
+ * @return whether an input is a TextInputType.
```
niedzielski, line 11: InputType

src/components/input/Input.snap.ts
Volker-E: @niedzielski's flag is correct. This shouldn't be the default for inputs.

src/components/input/Input.vue
```diff
@@ -0,0 +1,124 @@
...
+	name: 'WvuiInput',
+	inheritAttrs: false,
+	props: {
+		type: {
+			type: String as PropType<InputType>,
+			default: InputType.search,
+			validator: isInputType
+		},
+		disabled: {
+			type: Boolean as PropType<boolean>,
```
niedzielski, line 31: Do we need the PropType for Booleans too? I was just using it for the custom properties like InputType.

src/components/input/InputType.ts
```diff
@@ -0,0 +1,17 @@
+/*
+* Defines types for text inputs
+* */
```
niedzielski, line 3: Extra star here.

src/components/input/Input.vue
speedandfunction-anribolon: Yep, thanks.

PR #47, opened 2020-07-10
[update] Add initial icon component

AnneTee:
👋 🧩 

This is the icon component we built for Machine Vision for use in Button and Message components. In its current state, it uses OOUI classes for the icon background image, so it can only be used if OOUI icon images and `background-image` rules are available, e.g. in a Resource Loader module that requires them. This was fine for our purposes but obviously isn't as agnostic as this library is meant to be.

So, is this acceptable as a first iteration, or do we want to go ahead and add icon images and `background-image` styles? I wasn't sure if those were meant to be stored in this library. A temporary solution would be to include an asset or two for use in storybook.

If this gets merged, I can add our Message component and update the Button component to enable icons.

P.S. I based this on the `update/button-styles` branch so tests would pass, can switch this to `master` if needed


src/components/icon/Icon.vue
```diff
@@ -0,0 +1,64 @@
+<template>
+	<span class="wvui-icon" :class="classes">
```
AnneTee, line 2: Small thing: I noticed that y'all are including static class names in the dynamic `classes` computed property. One advantage to keeping them in the template is readability, especially for components that have child elements. All of your block- and element-level classes are listed in the template. Example in our Button component:

```
<template>
	<button
		class="mw-button"
		:class="classes"
		:disabled="disabled"
		@click="$emit( 'click' )"
	>
		<icon
			v-if="icon"
			:icon="icon"
			:invert="invert"
		></icon>
		<div class="mw-button__content">
			<slot />
		</div>
	</button>
</template>
```

The downsides:
- `class` and `:class` might seem redundant
- There are now 2 places in the code where class names are listed

The alternatives:
- Create computed properties for all class names (which would be kind of silly if none of them were actually computed)
- Only use `class` without v-bind for elements that only have static class names (which means your block- and element-level classes would appear in 2 different places)

santhoshtr:
We need to think what are the pain points in OOUI based icon approach and try to solve them by taking advantage of browser capability progress.

1. SVG support is now nearly [complete](https://caniuse.com/#search=svg) in browsers 
2. Treeshaking support allows us to ship only the required icons 
3. SVG icons allows to have one icons that use/inherit colors from the context instead of having a file for each colored icons.
4. We need to think how we are going to do icon flipping with in a context of RTL script

Alternate approach present in Content Translation for reference:  [Icon component](https://github.com/wikimedia/mediawiki-extensions-ContentTranslation/blob/master/app/src/lib/mediawiki.ui/components/MWIcon.vue), [icons](https://github.com/wikimedia/mediawiki-extensions-ContentTranslation/blob/master/app/src/lib/mediawiki.ui/components/icons.js) , [storybook](https://people.wikimedia.org/~santhosh/storybook/?path=/story/components--icons)

Volker-E:
> We need to think what are the pain points in OOUI based icon approach and try to solve them by taking advantage of browser capability progress.
>     4. We need to think how we are going to do icon flipping with in a context of RTL script

Automated flipping is not a proper solution from i18n/design perspective. Flipping for RTL makes only sense in the smallest amount of icons. It's intended to go through design and possibly native speakers review for each icon, as it's already in place for all Design Style Guide OOUI/WikimediaUI theme icons.

santhoshtr:
> > 4. We need to think how we are going to do icon flipping with in a context of RTL script
> 
> Automated flipping is not a proper solution from i18n/design perspective. Flipping for RTL makes only sense in the smallest amount of icons. It's intended to go through design and possibly native speakers review for each icon, as it's already in place for all Design Style Guide OOUI/WikimediaUI theme icons.

Sorry, I was not suggesting automatic flipping for all icons. I was suggesting the possibility of a smart Vue Icon component that abstract the flippable or not flippable property and just gives a valid icon rendering in a given context(script directionality, language, color, container size, variant icons). For a feature developer something like `<mw-icon icon="trash"/>` should get an icon in easy way. The framework should take care of getting the required assets to browser. I was showing an [example](https://people.wikimedia.org/~santhosh/storybook/?path=/story/components--icons) implementation in CX. This is not new, it is inspired by icon systems like [mdi/js](https://www.npmjs.com/package/@mdi/js), [mdi/svg](https://www.npmjs.com/package/@mdi/svg) and their usage with popular Vue UI frameworks([example](https://vuetifyjs.com/en/components/icons/)).

src/components/icon/Icon.vue
niedzielski: Thanks for pointing this out! I agree with your assessment of the pros and cons but weigh them differently. That is, I think that colocating the classes in a single place (either all statically, if possible, or all dynamically, if not) is the most readable. Perhaps an extreme example, If I had three components with dynamic classes, I wouldn't want to jump between six places.

All that said, I have so very little Vue experience and a quick review suggests this pattern is _common_. So, assuming no one else objects, I'm down for applying this suggestion across all of WVUI if it seems like a good idea to you. If so, please add the convention to the readme and I'll update my button patch. If not, please update the readme anyway with the approach you end up taking :laughing: Thanks again for taking the time to lay it out so neatly :+1:

src/components/icon/Icon.vue
```diff
@@ -0,0 +1,64 @@
...
+		{{ label }}
```
niedzielski, line 3: Did you consider using a slot here? Is it possible to pass in an emphasized label like `<em>Ba</em>nana`? Is label just for screen readers? If not, do you think that mixing icon and label in one component is too many responsibilities?

src/components/icon/Icon.vue
```diff
@@ -0,0 +1,64 @@
...
+	name: 'WvuiIcon',
+	props: {
+		icon: {
+			type: String,
+			required: true
+		},
+		invert: {
+			type: Boolean
+		},
+		label: {
```
niedzielski, line 25: Is this for screen readers only or also for presentation? Please consider adding a comment.

egardner:
> The framework should take care of getting the required assets to browser. I was showing an [example](https://people.wikimedia.org/~santhosh/storybook/?path=/story/components--icons) implementation in CX. This is not new, it is inspired by icon systems like [mdi/js](https://www.npmjs.com/package/@mdi/js), [mdi/svg](https://www.npmjs.com/package/@mdi/svg) and their usage with popular Vue UI frameworks([example](https://vuetifyjs.com/en/components/icons/)).

It seems like updating this component to use SVG icons is really worth considering here, especially if many (most? all?) of the required assets already exist in vector form. Is there any reason this wouldn't be feasible at this time?

src/components/icon/Icon.vue
egardner: I just want to say I think splitting static `class` and dynamic (computed-property-based) `v-bind:class` attributes is a good practice in Vue. 

I think that in single-file components, the proper place for things that don't change is in the template. This should also make it easier for folks like designers (who may not be as fluent in JS/Typescript) to look quickly at the HTML in order to understand the basics of what's going on. 

Finally, in Vue 3 at least, the template compiler is able to optimize code generated from templates to pretty granular level; static parts of a component's template get "hoisted" out and aren't constantly re-evaluated, so there may be some marginal performance improvements of this approach as well.

src/components/icon/Icon.vue
AnneTee: The label is only meant for screen readers and will always be hidden by CSS in the component's current state, and I do think that we should separate the icon component from a visible label (although I could be convinced otherwise if that's something we really need). I'd think that HTML in the visually-hidden label would be uncommon, but we could use a slot here to enable that.

src/components/icon/Icon.vue
AnneTee: Yes, although I think we should swap this out for a slot to allow HTML even if it's rarely needed, so this prop can be removed.

src/components/icon/Icon.vue
```diff
@@ -0,0 +1,64 @@
...
+			// String or message object.
+			type: [ String, Object ],
+			default: null
+		}
+	},
+	computed: {
+		classes(): Record<string, boolean> {
+			// Use the existing OOUI classes for icon images for now.
```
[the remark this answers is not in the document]
AnneTee: This was definitely only ever meant to be a temporary bandaid and I'm all for solving this problem now! Will comment more in the main thread...

src/components/icon/Icon.vue
AnneTee: Thanks for the comments! Given that this has several benefits and is a common pattern I say we go with it. I'll document it in the readme.

src/components/icon/Icon.vue
```diff
@@ -0,0 +1,64 @@
...
+
+.wvui-icon {
+	background-size: contain; // stylelint-disable-line plugin/no-unsupported-browser-features
+	background-position: center center;
+	background-repeat: no-repeat;
+	display: inline-block;
+	height: @size-icon / @font-size-browser / @font-size-base;
+	line-height: 2.5;
+	margin: 0;
+	min-height: @size-icon;
```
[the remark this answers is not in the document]
AnneTee: done!

src/components/icon/Icon.vue
```diff
@@ -0,0 +1,64 @@
...
+			};
+		}
+	}
+} );
+</script>
+
+<style lang="less">
+@import ( reference ) '@/themes/wikimedia-ui.less';
+
+.wvui-icon {
```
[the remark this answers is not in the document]
AnneTee: I added a few comments but mostly removed styles that came from OOUI if they met these criteria:

1. We don't currently need them in WVUI
2. I couldn't figure out why they would be necessary in WVUI in the future as we add more components

src/components/icon/Icon.stories.ts
```diff
@@ -0,0 +1,14 @@
+import Vue from 'vue';
+import WvuiIcon from './Icon.vue';
+
+export default {
+	title: 'Components/Icon',
+	parameters: { layout: 'centered' }
+};
+
+export const tag = (): Vue.Component => Vue.extend( {
```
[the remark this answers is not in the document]
AnneTee: Good call, done!

src/components/icon/Icon.stories.ts
```diff
@@ -0,0 +1,23 @@
+import { boolean, text } from '@storybook/addon-knobs';
+import Vue from 'vue';
+import WvuiIcon from './Icon.vue';
+
+export default {
+	title: 'Components/Icon',
```
niedzielski, line 6: I used the HTML tag here like `Components/wvui-button`. My thinking was that library developers will be sick to death of seeing WVUI-this and WVUI-that but library consumers might expect a leading `wvui-` when copying, pasting, and searching. I don't feel strongly about this point but it would be nice to be consistent one way or the other. Let me know your thoughts and we can update as appropriate.

src/components/icon/Icon.vue
niedzielski: Thank you!

src/components/icon/Icon.vue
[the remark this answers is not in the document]
niedzielski: Thanks!

src/components/icon/Icon.vue
```diff
@@ -0,0 +1,66 @@
...
+	background-size: contain; // stylelint-disable-line plugin/no-unsupported-browser-features
+	background-position: center center;
+	background-repeat: no-repeat;
+	box-sizing: border-box;
+	display: inline-block;
+	height: @size-icon / @font-size-browser / @font-size-base;
```
niedzielski, line 51: Can you add a comment as to why @size-icon isn't used?

src/components/icon/Icon.stories.ts
AnneTee: Using the prefix when a component is in use but not when it's being described, as we're currently doing, seems to be a common pattern in other libraries. Using `wvui` in the storybook tab names would feel like clutter IMO. Since it's a common pattern and we have the story available for each component demonstrating the prefix in use, I think it's okay to omit `wvui` here. I also don't feel that strongly about this other than maintaining consistency across this library!

src/components/icon/Icon.vue
AnneTee: Done 👍

src/components/icon/Icon.test.ts
```diff
@@ -0,0 +1,87 @@
...
+/* eslint-disable @typescript-eslint/no-explicit-any */
+it( 'handles ltr text direction', () => {
+	// Default dir is 'ltr'.
+	const wrapper = shallowMount( WvuiIcon, {
+		propsData: {
+			icon: iconVariesPerDir
+		}
+	} );
+
+	expect( ( wrapper.vm as any ).iconPath ).toMatch( iconVariesPerDir.paths.ltr );
```
AnneTee, line 53: I had to disable the `@typescript-eslint/no-explicit-any` rule to get this through, but that obviously doesn't feel right. Can anyone tell me what I should be doing here instead of using `any`?

AnneTee:
@niedzielski et al, I've pushed some updates that I'm betting will open up more discussion than they've resolved! Please see the commit messages for details. Some comments and questions:

- **Icon data format**: I based this off of CX (thanks again @santhoshtr), in which icon data can either be a string representing the svg path or an object with a path property (plus other data). I took this a step further in order to support icons that have different paths per text direction (so they can't just be flipped) and icons that have different paths per language. One benefit of the way it's set up is that the format is simple for most icons (just a string) and only needs to be longer/more complicated for one-offs and exceptions. This would make it easier for users to set up custom icons to use with the `WvuiIcon` component. However, the icon data feels very messy. Should I be taking advantage of TypeScript here? Any advice here is welcome (I'm new to TS)
- **Icon groups**: I thought it would make sense to allow developers to import (or create an entry for) individual icons or groups of icons, like we have with OOUI. Thoughts?
- **svg paths**: Since the Icon component uses a single path for the svg, I had to convert our existing icons to this format. The resulting svg icons could all use a close review at some point (cc @Volker-E)

src/components/icon/Icon.test.ts
niedzielski: Sadly, [I don't think there's a good way around this one](https://github.com/vuejs/vue-test-utils/issues/255) in Vue 2. The type of all components is currently limited to "generic Vue component"--see VueLoader.d.ts. For example, even if you do explicit typing instead of inference:

```ts
	const wrapper: WvuiIcon = shallowMount( WvuiIcon, {
		propsData: {
			icon: iconVariesPerLang
		}
	} ).vm
```

`WvuiIcon` is still of type `Vue` outside of WvuiIcon.vue. I'm not opposed to disabling this linter for the tests but it can be abused.

src/themes/icons.js
```diff
@@ -0,0 +1,549 @@
+/**
+ * These icon paths will be used by the Icon component to display an svg. Icons
+ * are not flippable by default.
+ *
+ * An icon can take the following forms:
+ *  - A path string
+ *  - An object containing a path property
+ *  - An object containing a paths property that contains different paths per
+ *    language (e.g. wvuiIconBold) or text direction (e.g. wvuiIconEditUndo).
```
santhoshtr, line 9: I don't have a strong opinion on how we structure them. But let me propose an alternate structure to choose from.

```javascript
// Simple icon, horizontally symmetric or flippable. Example: Add, Arrows
export const wvuiIconXyz= {
  path: 'M11 9V4H9v5H4v2h5v5h2v-5h5V9z'
}
// Should not be flipped in any context. An example is Wikipedia logo
export const wvuiIconXyz= {
  path: 'M11 9V4H9v5H4v2h5v5h2v-5h5V9z',
  flippable: false
}
// Icon has LTR and RTL script direction variants
export const wvuiIconXyzLTR= {
  path: 'M11 9V4H9v5H4v2h5v5h2v-5h5V9z'
}
export const wvuiIconXyzRTL = {
  path: 'M11 9V4H9v5H4v2h5v5h2v-5h5V9z'
}
export const wvuiIconXyz= {
  variants:{
     ltr:  wvuiIconXyzLTR,
     rtl:  wvuiIconXyzRTL
   }
}

export const wvuiIconXyzHi = {
  path: 'M11 9V4H9v5......5h2v-5h5V9z'
}
export const wvuiIconXyzFa = {
  path: 'M11 9V4H9v5H4v2.....h5v5h2v-5h5V9z'
}
// Icon has language variants
export const wvuiIconXyz= {
  variants:{
     // LanguageCode: IconObject
     hi:  wvuiIconXyzHi
     fa:  wvuiIconXyzFa
   }
  // default path
   path:  'M11 9V4H9v5H4v2h5v5h2v-5h5V9z'
}
```

This make sure that svg paths are always mapped to an Icon name and not to language codes, arbitrary variant names or ltr/rtl. This also allows to explicitly render an icon variant - may be for a demo purpose. Since an icon object now has a recursive and modular structure, we are using a minimal structure- 3 possible members: path, flippable, variants

src/components/icon/Icon.stories.ts
```diff
@@ -0,0 +1,83 @@
...
+			return {
+				iconGroups,
+				icon: null
+			};
+		},
+		beforeMount() {
+			window.document.documentElement.setAttribute( 'dir', 'ltr' );
+		},
+		template: `
+		<div>
```
santhoshtr, line 28: Both LTR and RTL can be combined into single story. if we add `<div :dir="scriptDirection">` and scriptDirection being  a property(changeable using a knob). See https://people.wikimedia.org/~santhosh/storybook/?path=/story/components--icons for an example

src/components/icon/Icon.vue
```diff
@@ -0,0 +1,121 @@
...
+		},
+		/** The height and width of the icon, in pixels. */
+		size: {
+			type: [ Number, String ],
+			default: 20
+		}
+	},
+	data(): Record<string, string> {
+		return {
+			langCode: document.documentElement.lang,
```
santhoshtr, line 62: language need to be an optional property to allow element blocks with multiple lanugages co-exist in a document. For  content translation usecase, this is very important -we often have content from two different languages in same screen. So the parent components should be able to pass it to this component as contextual variable. This will also help you to write a storybook demoing language variant capability of icons :-). It an ofcourse fallback to documentElement.lang

src/themes/iconGroups.js
```diff
@@ -0,0 +1,260 @@
+/**
```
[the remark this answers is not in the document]
AnneTee: These are all excellent points @santhoshtr, thanks for explaining your thoughts.

From a development and code bundling standpoint, the groups may be obsolete now that we can use tree-shaking or, for circumstances where that's not possible (e.g. a Structured Data project where we're exclusively using Resource Loader), create project- or feature-specific entries that define the required icons (something Stephen has suggested).

I agree that we should think about optimizing the developer experience of finding and choosing an icon, or of remembering an icon name and using it without looking it up. When coding, it's obviously easier to remember an icon name rather than icon group + icon name. One of the reasons I ported over the icon groupings from OOUI was to maintain consistency between the [OOUI icon demo page](https://doc.wikimedia.org/oojs-ui/master/demos/?page=icons&theme=wikimediaui&direction=ltr&platform=desktop) and the new Storybook demo. I'm assuming the groupings were carefully chosen and that it's optimal for a designer or developer to search through icons grouped by categories rather than, say, an alphabetical list. That said, I don't have much of an opinion here and am very open to changing the groups, removing them from everything except Storybook, or removing them altogether.

@Volker-E, thoughts?

src/themes/icons.js
```diff
@@ -0,0 +1,549 @@
+/**
+ * These icon paths will be used by the Icon component to display an svg. Icons
+ * are not flippable by default.
```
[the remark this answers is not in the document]
AnneTee: I think I should change that property to `shouldFlip` instead of flippable. My reasons for defaulting to false (really, null) are that we're following that paradigm in other places (e.g. component props), it allows us to simplify the CSS a little bit, and it simplifies the logic for adding the flippable/should-flip class a little bit. "Flippable" seemed confusing to me since it doesn't address icons that are different per text direction but can't just be flipped, and because we don't need to pay attention to symmetric icons when thinking about all this.

Does changing the terminology here solve the problem?

src/components/icon/Icon.stories.ts
AnneTee: @santhoshtr I had it set up this way originally, where changing the knob would update the `dir` attribute on the document element within the iframe, but it didn't work for the icons that have different paths per text direction because the Icon component isn't reactive to changes in those HTML attributes. I think the solution is to use the Storybook context add-on, but I didn't dig into this much due to lack of time. It would be much nicer if we could switch text direction via a knob. I'll try to get it working if I can, but it might have to be a future improvement.

AnneTee: Also.....do we need the Icon component to be reactive to changes in dir or lang? I didn't think so with the way language switching happens onwiki, but please let me know if you disagree!

src/components/icon/Icon.vue
AnneTee: That makes sense, thanks for the info! I'll make this change and update the story so we can demo this.

Volker-E:
I swear by all things accountable I haven't closed this willingly or knowingly.

AnneTee:
@Volker-E I thought this was your subtle way of telling me you don't approve of this PR :joy:

But seriously, maybe it was closed automatically when the branch it was opened against was merged into master. I'll reopen it and change the base branch!

niedzielski:
Hi @AnneTee, I asked @Volker-E to merge the button patch which closed that branch and I guess killed this. When I hit edit it won't let me change the branch. Can you reopen and edit?

AnneTee:
@niedzielski I actually can't reopen it, maybe someone who's an admin of this repo can?

Volker-E:
By all seriousness, clearly not my style. I've merged #42 at around the same time, so yes, that could (somewhat) reasonably be it. Although the GitHub UI could be a bit more obvious here. And tried to edit like @niedzielski without success. @AnneTee Big Sorry!

niedzielski:
I think this is my fault for configuring the repo to auto delete branches. I will update the setting and we can see how it goes. Sorry for the confusion!

AnneTee:
Yay, thanks @niedzielski!

Totally kidding @Volker-E, I figured it was unintentional 🙃

niedzielski:
I had to restore the button branch to reopen this and change its destination to master. I've disabled "Automatically delete head branches". Very sorry about this! This one's on me not @Volker-E.

src/components/icon/Icon.snap.ts
```diff
@@ -0,0 +1,111 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`matches the snapshot Case 0 With icon: ({"icon": "path string"}) => HTML 1`] = `
+<span
+  class="wvui-icon"
+>
+  <svg
+    aria-hidden="true"
```
Volker-E, line 8: If it's set to `aria-hidden` it doesn't need `role=presentation`. Thinking of dynamic modification or certain use cases, where it will be overriden?

src/themes/icons.js
```diff
@@ -0,0 +1,549 @@
+/**
```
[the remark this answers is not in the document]
AnneTee: Thanks Stephen, based on your comment and the one from @santhoshtr below, I've added a few icon types, standardized the data structure, and simplified language and text dir maps, see [here](https://github.com/wikimedia/wvui/pull/47/commits/bed72bba26c82d2e5871268cc7a5bf188cd94569#diff-aa394e4d23a38bce8cdf5a87b48449ec).

src/themes/icons.js
AnneTee: This was really helpful! I've implemented something very similar, feedback welcome.

src/themes/iconGroups.js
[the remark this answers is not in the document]
Volker-E: The grouping came first and foremost from a technical limitation in my knowledge: In order to get them delivered in a reasonably performant way as RL modules, they were grouped. It could be that, for some of the packs, like editing-* or icons-wikimedia it also gave designers a simpler orientation, but this seems a secondary thought here, that we don't need to follow. Looking at the arbitrary boundary between 'editing-list' and 'editing-styling' for example.

Volker-E:
> We need to think what are the pain points in OOUI based icon approach and try to solve them by taking advantage of browser capability progress.
> 
>     1. SVG support is now nearly [complete](https://caniuse.com/#search=svg) in browsers
> 
>     2. Treeshaking support allows us to ship only the required icons
> 
>     3. SVG icons allows to have one icons that use/inherit colors from the context instead of having a file for each colored icons.

I think this approach is neglecting one detail. We would send the SVG icon repeatedly down the wire in more complex interfaces where we repeat the icon. Think a close icon in many tag items like in Growth teams or the longer Notifications lists with repeated icons in Echo. Doesn't seem to be a stopper here on GitHub (looking at the interface around). Nonetheless worth mentioning. We currently rely on the same cached icon with CSS `background-image` through RL.

src/components/icon/Icon.vue
```diff
@@ -0,0 +1,127 @@
+<template>
+	<span class="wvui-icon" :class="classes">
+		<svg
```
Volker-E, line 3: Boah, that way of writing HTML will drive me nuts long-term. 24 years of writing HTML being overthrown by Vue. Please @niedzielski let's revisit this.

src/components/icon/Icon.vue
```diff
@@ -0,0 +1,127 @@
...
+			xmlns="http://www.w3.org/2000/svg"
+			width="1em"
+			height="1em"
+			viewBox="0 0 20 20"
+			aria-hidden="true"
+			role="presentation"
```
Volker-E, line 9: Stated above, `role=presentation` without value when `aria-hidden` is set too.

src/components/icon/Icon.vue
```diff
@@ -0,0 +1,127 @@
...
+		>
+			<g :fill="iconColor">
+				<path :d="iconPath" />
+			</g>
+		</svg>
+		<span class="wvui-icon__content"><slot /></span>
```
Volker-E, line 15: When we're inlining SVG markup, we should actually rely on SVG's `title` tag instead. 
Currently I can't think of a usage of icons within OOUI where an inlined SVG with `title` wouldn't work better than the combination of SVG without `title` and added `span`. It would be the most forward, future-proof solution and save us a DOM element every Icon application on top.
Note that I need to go through the screenreader support, current OOUI widgets and readout once more before that's fully clarified.

src/components/icon/Icon.vue
```diff
@@ -0,0 +1,127 @@
...
+ * language. This component relies on the dir and lang attributes of the <html>
+ * element, respectively, to handle those instances. See src/themes/icons.ts
+ * for data structure (e.g. wvuiIconBold).
+ *
+ * Slot may contain text for screen readers and will be visually hidden.
+ */
+export default Vue.extend( {
+	name: 'WvuiIcon',
+	props: {
+		/** The svg path or an object containing that path plus other data. */
```
Volker-E, line 44: Minor: “SVG”

src/components/icon/Icon.vue
```diff
@@ -0,0 +1,127 @@
...
+	},
+	data(): Record<string, string> {
+		return {
+			dir: document.documentElement.dir || 'ltr'
+		};
+	},
+	computed: {
+		classes(): Record<string, boolean> {
+			return {
+				'wvui-icon--should-flip': this.shouldFlip
```
Volker-E, line 68: Sounds like a SMACSS state class, but is it? http://smacss.com/book/type-state
We should probably consider naming alternatives here…

src/components/icon/Icon.stories.ts
santhoshtr: For CX, we do change the languages all the time(on the fly), and reactivity was required.

src/themes/icons.js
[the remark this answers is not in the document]
santhoshtr: I was also thinking whether you used "Should flip" logic in the code. No issuse is renaming to it. Just make sure to document in icons.js and wherever necessary. This could be a difficult to understand flag for future developers.

santhoshtr:
> I think this approach is neglecting one detail. We would send the SVG icon repeatedly down the wire in more complex interfaces where we repeat the icon. 

@Volker-E As per the latest approach used here, SVGs are going to be inside a javascript file, compiled by the build step. SVG paths alone will be there, not the full svg file. This javascript file will have gone through tree shaking - unwanted icons/unreferenced icons wont be going to build output. So there is no SVG file request separate from this one time request for javascript file managed by RL.

src/components/icon/Icon.stories.ts
AnneTee: Got it, and that'll work here now that `langCode` is a prop

src/components/icon/Icon.vue
AnneTee: Good point, maybe `wvui-icon--flip-for-rtl` or just `wvui-icon--flip`?

src/themes/iconGroups.js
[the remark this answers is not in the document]
AnneTee: Cool, I'll remove the groups then. Thanks!

src/components/icon/Icon.vue
AnneTee: This makes way more sense; I've made this update [here](https://github.com/wikimedia/wvui/pull/47/commits/8c9101002d057bb9fb893a17e28ae4bbf0052e72)

src/components/icon/Icon.stories.ts
niedzielski: I just ended up using what you have here for the button ('Components/Button').

src/components/icon/Icon.vue
niedzielski: I don't wish to misinterpret. Can you help me understand what your concerns are?

src/themes/icons/iconTypes.ts
```diff
@@ -0,0 +1,34 @@
+/**
+ * An icon with a single path.
+ */
+export interface Icon {
```
niedzielski, line 4: This seems very similar to IconVariedByDir. I don't think it matters much either way but maybe the two could be merged if you think it's clearer:

```ts
export interface Icon {
  /** The default SVG path. Use this for all LTR languages. When rtlPath is unspecified and for exceptions, this is also the icon to be used. */
  path: string,
  /** The SVG path to use for right-to-left languages. If the language direction is RTL and not excepted below, this is the path to use. */
  rtlPath?: string,
  /** Right-to-left languages in this array are special cased to use the left-to-right icon. */
  rtlExceptions: string[]
}
```

This isn't a performance request. I'm just highlighting that the cases seemed similar and perhaps could be handled the same way if you think it's clearer. I go back and forth.

src/components/icon/Icon.vue
```diff
@@ -0,0 +1,132 @@
...
+ * language. This component relies on the dir and lang attributes of the <html>
+ * element, respectively, to handle those instances. See src/themes/icons.ts
+ * for data structure (e.g. wvuiIconBold).
+ *
+ * Slot may contain text for screen readers and will be visually hidden.
+ */
+export default Vue.extend( {
+	name: 'WvuiIcon',
+	props: {
+		/** The SVG path or an object containing that path plus other data. */
```
niedzielski, line 43: Is this expected to take an Icon type like the one below? If so, please mention that in the docs here or use a prop type something like:

```
import Vue, { PropType } from 'vue';
import { Icon, IconVariedByLang, IconVariedByDir } from '../../themes/icons/iconTypes';

// Consider adding this to iconTypes
type AnyIcon = string | Icon | IconVariedByLang | IconVariedByDir

...
		icon: {
			type: [ String, Object ] as PropType<AnyIcon>,
			required: true
		},
```

src/components/icon/Icon.vue
```diff
@@ -0,0 +1,132 @@
...
+		},
+		/**
+		 * Accessible title for SVG. String or message object. If not included,
+		 * the SVG will be hidden from screen readers.
+		 */
+		iconTitle: {
+			type: [ String, Object ],
+			default: ''
+		},
+		/** Explicitly set the language or default to document lang. */
```
niedzielski, line 61: It might be useful to clarify that any valid HTMLElement.lang is valid here if that's true and link to the MDN docs.

https://developer.mozilla.org/docs/Web/API/HTMLElement/lang

src/components/icon/Icon.vue
```diff
@@ -0,0 +1,132 @@
...
+	},
+	computed: {
+		rootClasses(): Record<string, boolean> {
+			return {
+				'wvui-icon--should-flip': this.shouldFlip
+			};
+		},
+		shouldFlip(): boolean {
+			// Don't flip if the current language is listed as an exception.
+			if ( this.icon.shouldFlipExceptions ) {
```
niedzielski, line 80: I believe you've written this below as an always defined array so that you don't need to check whether it is undefined or null here.

src/components/icon/Icon.vue
```diff
@@ -0,0 +1,132 @@
...
+				return this.icon.shouldFlipExceptions.indexOf( this.langCode ) === -1;
+			}
+
+			return this.icon?.shouldFlip;
```
niedzielski, line 84: I didn't realize that the tree-shaking typechecking applied to the more general case of all library code, not just the configuration files. That is, Icon is not exported by any entry so it is not being type checked. Please add it to the src/entries/wvui.ts and consider revising the known issues listed in the readme with something like "_Dead code and_ JavaScript configuration files are not type checked..."

I'm not sure I have the logic right but by using a PropType, you should be able to get some type checking in:

```ts
		shouldFlip(): boolean {
			if ( typeof this.icon === 'string' ) {
				return true;
			}

			if ( 'shouldFlipExceptions' in this.icon ) {
				// Don't flip if the current language is listed as an exception.
				const exception = this.icon.shouldFlipExceptions?.indexOf( this.langCode );
				return exception === undefined || exception === -1;
			}

			if ( 'shouldFlip' in this.icon ) {
				return !!this.icon.shouldFlip;
			}

			return true;
		},
```

niedzielski: And, if you think it's a good idea, consider moving this to a utility function in iconTypes.ts like:

```ts
export function shouldFlip(icon: AnyIcon): boolean {
  ...
}
```

This might make it easier to test and allow you to focus on UI here in the component.

AnneTee: Done, and I moved the getIconPath logic to the types file as well.

src/components/icon/Icon.vue
AnneTee: This should be resolved now that the code has been refactored

src/components/icon/Icon.vue
AnneTee: Done

src/components/icon/Icon.vue
AnneTee: Done

src/themes/icons/iconTypes.ts
AnneTee: I opted to keep IconVariedByDir distinct for a couple of reasons:

- `IconVariedByDir` is not meant to use the `shouldFlip` property, which could be confusing if it were merged with `Icon`
- I like the consistency of `IconVariedByLang` and `IconVariedByDir` as icon sets that list `Icon` components for use in different circumstances

That said, I'm flexible on the shape of the icon types if there's a better or simpler way to set them up!

src/themes/icons/iconTypes.ts
```diff
@@ -0,0 +1,34 @@
...
+	 */
+	shouldFlipExceptions?: Array<string>
+}
+
+/**
+ * An icon that varies per language.
+ */
+export interface IconVariedByLang {
+	/** Language code key with corresponding icon. */
+	langVariants: Record<string, Icon>,
```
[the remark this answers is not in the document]
AnneTee: Whew, this was hard to name! I agree that "variants" is the wrong word and went for `langCodeMap`. Suggestions for improvements welcome if anyone has a different idea!

src/themes/icons/iconTypes.ts
```diff
@@ -0,0 +1,34 @@
...
+	/** The default icon. */
+	default: Icon
+}
+
+/**
+ * An icon that varies per text direction (but can't just be flipped).
+ */
+export interface IconVariedByDir {
+	/** Text direction key with corresponding icon. */
+	dirVariants: Record<string, Icon>,
```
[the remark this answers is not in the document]
AnneTee: Done, good call!

src/themes/icons/iconTypes.ts
```diff
@@ -0,0 +1,34 @@
...
+export interface Icon {
+	/** SVG path. */
+	path: string,
+	/** Whether the icon should be flipped horizontally via CSS in RTL mode. */
+	shouldFlip?: boolean,
+	/**
+	 * Language codes that are exceptions to the above property (e.g. the help
+	 * icon should flip in RTL mode, but not for Hebrew or Yiddish).
+	 */
+	shouldFlipExceptions?: Array<string>
```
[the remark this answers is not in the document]
AnneTee: Done

src/themes/icons/icons.ts
```diff
@@ -0,0 +1,942 @@
+/**
```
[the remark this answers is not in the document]
AnneTee: This documentation was definitely redundant and I've removed most of it.

src/themes/icons/icons.ts
```diff
@@ -0,0 +1,942 @@
...
+ * These icon paths will be used by the Icon component to display an svg. Icons
+ * are not shouldFlip by default.
+ *
+ * An icon can take the following forms:
+ *  - A path string
```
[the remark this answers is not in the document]
AnneTee: Done

src/themes/icons/icons.ts
```diff
@@ -0,0 +1,942 @@
...
+	path: 'M5 1a2 2 0 00-2 2v14a2 2 0 002 2h10a2 2 0 002-2V3a2 2 0 00-2-2zm0 3h5v1H5zm0 2h5v1H5zm0 2h5v1H5zm10 7H5v-1h10zm0-2H5v-1h10zm0-2H5v-1h10zm0-2h-4V4h4z',
+	shouldFlip: true
+};
+export const wvuiIconArticleAdd: Icon = {
+	path: 'M5 1c-1.1 0-2 .9-2 2v14c0 1.1.9 2 2 2h10c1.1 0 2-.9 2-2V3c0-1.1-.9-2-2-2zm10 10h-4v4H9v-4H5V9h4V5h2v4h4z'
+};
+export const wvuiIconArticleCheck: Icon = {
+	path: 'M9 17l-4.59-4.59L5.83 11 9 14.17l8-8V3a2 2 0 00-2-2H5a2 2 0 00-2 2v14a2 2 0 002 2h10a2 2 0 002-2V9z'
+};
+export const wvuiIconArticleDisambiguation = {
```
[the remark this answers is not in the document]
AnneTee: Done

src/themes/icons/icons.ts
```diff
@@ -0,0 +1,942 @@
...
+export const wvuiIconHistory: Icon = {
+	path: 'M9 6v5h.06l2.48 2.47 1.41-1.41L11 10.11V6z M10 1a9 9 0 00-7.85 13.35L.5 16H6v-5.5l-2.38 2.38A7 7 0 1110 17v2a9 9 0 000-18z'
+};
+export const wvuiIconHome: Icon = {
+	path: 'M10 1L0 10h3v9h4v-4.6c0-1.47 1.31-2.66 3-2.66s3 1.19 3 2.66V19h4v-9h3L10 1z'
+};
+export const wvuiIconImage: Icon = {
+	path: 'M2 2a2 2 0 00-2 2v12a2 2 0 002 2h16a2 2 0 002-2V4a2 2 0 00-2-2zm-.17 13l4.09-5.25 2.92 3.51L12.92 8l5.25 7z'
+};
+export const wvuiIconImageAdd = {
```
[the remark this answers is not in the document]
AnneTee: Done

src/themes/icons/icons.ts
```diff
@@ -0,0 +1,942 @@
...
+};
+export const wvuiIconSettings: Icon = {
+	path: 'M11.5 0l.42 2.75a7.67 7.67 0 0 1 1.87.77L16 1.87 18.16 4 16.49 6.23a7.67 7.67 0 0 1 .76 1.85L20 8.5v3l-2.75.42a7.67 7.67 0 0 1-.77 1.87L18.13 16 16 18.16l-2.24-1.67a7.67 7.67 0 0 1-1.85.76L11.5 20h-3l-.42-2.75a7.45 7.45 0 0 1-1.86-.77L4 18.13 1.87 16l1.65-2.23a7 7 0 0 1-.77-1.85L0 11.5v-3l2.75-.42a7.45 7.45 0 0 1 .77-1.86L1.87 4 4 1.87 6.23 3.52a7.17 7.17 0 0 1 1.85-.77L8.5 0ZM10 6.5A3.5 3.5 0 1 0 13.5 10 3.5 3.5 0 0 0 10 6.5Z'
+};
+export const wvuiIconSignature = {
```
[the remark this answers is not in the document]
AnneTee: Done

src/components/icon/Icon.vue
```diff
@@ -0,0 +1,132 @@
...
+ */
+export default Vue.extend( {
+	name: 'WvuiIcon',
+	props: {
+		/** The SVG path or an object containing that path plus other data. */
+		icon: {
+			type: [ String, Object ],
+			required: true
+		},
+		/** Numerical color value (e.g. hex code, rgba) or keyword. */
```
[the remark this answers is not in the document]
AnneTee: This is a [paint value](https://www.w3.org/TR/SVG/painting.html#SpecifyingPaint) – I'll add more info to the doc

src/components/icon/Icon.vue
```diff
@@ -0,0 +1,132 @@
...
+ * SVG icon.
+ *
+ * See src/themes/icons.ts for a list of all icons and src/themes/iconGroups.ts
+ * for icons grouped by category. To use an icon, import it or the iconGroup
+ * it's in, assign it to a name in your component's data option, then use v-bind
+ * to set the icon attribute of the <wvui-icon> element to that name.
+ *
+ * Alternately, custom or third-party icons could be used as long as the icon
+ * prop provided to this component is either a string containing the icon path
+ * or an object with a `path` property.
```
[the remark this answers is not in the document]
AnneTee: I think this is covered now that I've added the proptype

AnneTee:
@niedzielski I've responded to most of your comments with the exception of a couple related to `dir` which I'm hoping Santhosh can weigh in on. I'll add the icons and the icon component to the main entry point, but I was wondering how you want to handle increasing the max size configurations as the main JS file inflates a lot when you add all the icons.

src/components/icon/Icon.vue
```diff
@@ -0,0 +1,110 @@
...
+	// Equivalent to @size-icon in ems relative to our base 0.875em font size.
+	font-size: unit( @min-size-icon ) / @font-size-browser / @font-size-base;
+	justify-content: center;
+	// For inline, inline-block, and table layouts.
+	vertical-align: middle;
+	user-select: none;
+}
+
+// Horizontally flip icons that should be flipped for RTL languages.
+[ dir='rtl' ] .wvui-icon--flip-for-rtl svg {
```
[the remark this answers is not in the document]
AnneTee: @santhoshtr good call, it's easy to grab the computed style when the component mounts (see [this commit](https://github.com/wikimedia/wvui/pull/47/commits/a9a6f4f06297bbde08ef093ab688bec2055f26f1)). Is this something we want to do with `lang` as well, or would you prefer the option of passing that in explicitly as a prop, as we currently have it set up?

src/themes/icons/iconTypes.ts
niedzielski: Not much addd except to say that everything is looking very, very good :+1:

src/components/icon/iconTypes.ts
```diff
@@ -0,0 +1,86 @@
+/**
+ * An icon with a single path.
+ */
+export interface Icon {
```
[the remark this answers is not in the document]
AnneTee: Yes! There are some icons of this type that should not flip for RTL, e.g. `play`.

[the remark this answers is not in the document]
niedzielski: Should that be a simple SVG path string type instead then? Those are no-flips, right?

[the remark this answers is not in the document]
AnneTee: Yeah, some of these could just be strings. When I first created the icon types I converted all icons to objects, but later added back in support for strings as an icon type. So, some of these can be converted back to strings.

Volker-E:
It seems to use a few outdated icons. Few is a guess, but I've already identified 'articlesSearch' and it came with a few other ones that were merged into OOUI master at the same time. Notice the page [background position difference](https://doc.wikimedia.org/oojs-ui/master/demos/?page=icons&theme=wikimediaui&direction=rtl&platform=desktop).

AnneTee:
Good catch @Volker-E, I had missed a few icons that differ between text directions but can't be flipped via CSS. I did another review of the RTL icons and believe I've caught them all now, but please do let me know if you see anything else I've missed. I also fixed a small issue with the book icon. Thanks for the feedback!

Other than potential items from my lastest commits that haven't been reviewed yet, I think the only outstanding thing is finalizing how we want to get `dir` and `langCode` and what the defaults should be. This is how it's currently set up:

- `dir`: default to the document direction, then try to get the value from the element's computed style in the mounted hook
- `langCode`: this is an optional prop that defaults to the document lang

2 questions that perhaps @santhoshtr should answer:
1. To avoid accessing the `document` global, should we drop the defaults to document dir and lang? It seems to me we can rely on the element's computed style instead.
2. Is there a use case for adding `dir` as a prop like `langCode`, or should we just get the `dir` from the computed style?

AnneTee:
Oops, forgot about indicators. It seems that 3 of the indicators are already included in the icon set, so I can just add the `required` icon if it's needed.

src/components/icon/Icon.snap.ts
niedzielski: Is this resolved?

src/components/icon/Icon.stories.ts
niedzielski: Was this resolved?

src/components/icon/iconTypes.ts
```diff
@@ -0,0 +1,95 @@
+type SvgPath = string
+
+/**
+ * An icon with a single path.
```
niedzielski, line 4: I guess this is covered by the function below but please consider documenting the flipping assumption (either these should be flipped in all RTL languages or these should never be flipped).

src/components/icon/Icon.snap.ts
Volker-E: Needs more inputs from @AnneTee…

AnneTee: I've removed `role="presentation"` completely. Icons will get `aria-hidden="true"` if no `iconTitle` prop is passed in. If an `iconTitle` is provided, a `<title>` element will be added to the SVG and the `aria-hidden` attribute will not be included.

src/components/icon/Icon.vue
Volker-E: Why do we have to indent and break attributes in static elements? It's really hard to get over different approach of writing HTML.

src/components/icon/Icon.vue
niedzielski: Is this resolved?

src/components/icon/Icon.vue
niedzielski: Elements and their attributes can be quite lengthy. I have a small laptop screen so I personally enjoy the line breaks and see similar perspectives in other tools like [Prettier](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeAhAEQPIGEAqAmgAoCiABABYwC2ANgHwA6UqAEvgLIAy5u3AQQDKQgLxMQUCAFoAVgGdyNQtNx0B8oROZRy5dqQGYdevak6l8AvmwEAlIZfEgArjABm0gBzaWp-fgAkvjcpAycAJ7kMMF0pKgA9EEhYX6m5pbWAHICFs4AJnDyYABOAJYADjBl0BLkkLAIMM6RfBBZMKRQ+L66+glshsZpqABGEPkRJunEDGxwdHQQ5ADuECV0+eioAEJ2DOT4lGWKJ+QcPACs5DsQZXRwJRV0AIbwAHSJs2lmQrh2gWI+Gm-hWZSg+QgK3eAHMXuRROR3C4oGBqtByAAKACU5GA5Dh7wAju8Ki55JRMS8SjCXDQmvJcQBfADcBJexIR5AA2gBdNmEuhcgDUUDgK3ImDecBZP1McMxAHJSnBpYqADTkRUAVQE0gAGobDdJCBqtS83BBFdiBS8lfIEPkzYqKi8YXAAG5lcXWuWJP4AoEg-QBwH4cjyEpgZzUGAVeRIBIJFYp2EQCAwh7SF5QF50CLVMDyd6QGgJHN5gtlIvvBR1YSELK4SWkABipH2-v+YemiXGkx0iQu3AYIHVIAgVRqUHkyFA1JKUOI1IQs5QeZWLwis-HoxKLzAAGs4DAhK6wOCYcgYCUXHBx9R6AB1Y7weTnuBCVdlapemARZBwHkHcQHBB0ShgYh9xhGgXmQdw8wdccFAADx2fcjxPIQXnpbhwTgeDEPvEBUKES8HgARRcCB4EIugkJAV0SnAwDHzoMdGPKWAnzKfIYEoZAvAABnHCpFwdJ99wqQCxKKR4PQI8ciWo+AoMnNcQBeeRpDFOBCnyDiSjgZSyiMqC3VguiGIdGgymvW9iPkci4ComiCKQBD6OImAXlGHi+IEpAACZxxvF57kvXAIBoSyUCKS4OPJOB8F8tdPIYj070CCEmiEUpKhgAQISEf8HisuAmSZIA). That's just my opinion, of course. I'm sure others have different perspectives.

src/components/icon/Icon.stories.ts
AnneTee: @niedzielski I didn't merge the text direction stories for the reasons mentioned in my comment above. `langCode` has been added as a prop as requested. I still have [some open questions](https://github.com/wikimedia/wvui/pull/47#issuecomment-666717345) about how we're handling `dir` and `lang`, but I think this particular comment is resolved.

src/components/icon/iconTypes.ts
AnneTee: Good point, I've added a comment to explain this!

src/components/icon/Icon.stories.ts
niedzielski: :+1: Cool, I'm going to close this particular thread. Please reopen as needed!

src/components/icon/Icon.snap.ts
Volker-E: Looks good to me!

src/components/icon/Icon.snap.ts
```diff
@@ -0,0 +1,98 @@
...
+    height="1em"
+    viewBox="0 0 20 20"
+    width="1em"
+    xmlns="http://www.w3.org/2000/svg"
+  >
+    <!---->
+     
+    <g
+      fill="currentColor"
```
Volker-E, line 17: I feel strongly that this might not help us all over, but hurt us in several cases. We often have very subtle icon color notes where the parent color should have nothing to say about the icon color itself.
When set to `currentColor` we're taking away possibility to make a greyish icon easily darker (for hover/focus states for example). I'm clear that it would enable more component-dynamic coloring, but that needs to be proven in real products.
We could also work with CSS `filter`s to overcome above limitations, but filter calculations are complex. 
I'd stick to `opacity` in components and black icons by default for now and make this a separate patch to better argue about the benefits and risks.

src/components/icon/Icon.stories.ts
```diff
@@ -0,0 +1,83 @@
...
+export default {
+	title: 'Components/Icon',
+	parameters: { layout: 'centered' }
+};
+
+export const iconsLTR = (): Vue.Component =>
+	Vue.extend( {
+		components: { WvuiIcon },
+		props: {
+			iconColor: { type: String, default: color( 'Icon color', 'rgba(32,33,34,1)' ) },
```
Volker-E, line 15: whitespaces in rgba pls :)

src/components/icon/Icon.stories.ts
```diff
@@ -0,0 +1,83 @@
...
+				icons,
+				icon: null
+			};
+		},
+		beforeMount() {
+			window.document.documentElement.setAttribute( 'dir', 'ltr' );
+		},
+		template: `
+		<div>
+			<div v-for="icon in Object.keys( icons )" :key="icon">
```
Volker-E, line 32: here we put the attribs on one line. I wanna see that for static HTML and SVG everywhere :)

src/components/icon/Icon.stories.ts
```diff
@@ -0,0 +1,83 @@
...
+				icon: null
+			};
+		},
+		beforeMount() {
+			window.document.documentElement.setAttribute( 'dir', 'rtl' );
+		},
+		template: `
+		<div>
+			<div v-for="icon in Object.keys( icons )" :key="icon">
+				<p>
```
Volker-E, line 70: Why wrapped in `<p>`?

src/components/icon/Icon.vue
Volker-E: How did you deal with HTML the last couple of years, where attributes auto-wrapped in next line? :P

src/components/icon/Icon.vue
```diff
@@ -0,0 +1,110 @@
+<template>
+	<span class="wvui-icon" :class="rootClasses">
+		<svg
+			xmlns="http://www.w3.org/2000/svg"
+			width="1em"
```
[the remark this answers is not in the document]
Volker-E: @santhoshtr what customizable icon sizes come to your mind? Design team has settled basically on two sizes: 20px and 12px. 
Mind, that current icon sizing proposed in this patch goes a different direction. Where I'm not sure if we understand the implications. I would pledge for excerpting these questions into separate Phab tasks, it's really hard to keep overview of the many moving parts in a patch like this with already 133 comments without my own now. Feel also bad for @AnneTee and @niedzielski to try to keep the box closed as good as it gets.

src/components/icon/Icon.snap.ts
```diff
@@ -0,0 +1,98 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`matches the snapshot Case 0 With icon: ({"icon": "path string"}) => HTML 1`] = `
+<span
+  class="wvui-icon"
+>
+  <svg
+    aria-hidden="true"
+    height="1em"
```
Volker-E, line 9: So `1em` and `viewbox` with its `20px` might result in side-effects. More a feeling than a test result (yet).

src/components/icon/Icon.vue
```diff
@@ -0,0 +1,116 @@
...
+		 * https://www.w3.org/TR/SVG/painting.html#SpecifyingPaint for all
+		 * options.
+		 */
+		iconColor: {
+			type: String,
+			default: 'currentColor'
+		},
+		/**
+		 * Accessible title for SVG. String or message object. If not included,
+		 * the SVG will be hidden from screen readers.
```
Volker-E, line 53: will be hidden from screen readers via `aria-hidden="true"`

src/components/icon/Icon.vue
```diff
@@ -0,0 +1,116 @@
...
+	align-items: center;
+	// Maintain an inline outer element while using flexbox to center the SVG
+	// and avoid extra space around the image.
+	display: inline-flex; // stylelint-disable-line plugin/no-unsupported-browser-features
+	// Equivalent to @size-icon in ems relative to our base 0.875em font size.
+	font-size: unit( @min-size-icon ) / @font-size-browser / @font-size-base;
+	justify-content: center;
+	// For inline, inline-block, and table layouts.
+	vertical-align: middle;
+	user-select: none;
```
Volker-E, line 109: Don't think it's necessary to set `user-select` here.

src/components/icon/Icon.vue
```diff
@@ -0,0 +1,116 @@
...
+.wvui-icon {
+	align-items: center;
+	// Maintain an inline outer element while using flexbox to center the SVG
+	// and avoid extra space around the image.
+	display: inline-flex; // stylelint-disable-line plugin/no-unsupported-browser-features
+	// Equivalent to @size-icon in ems relative to our base 0.875em font size.
+	font-size: unit( @min-size-icon ) / @font-size-browser / @font-size-base;
+	justify-content: center;
+	// For inline, inline-block, and table layouts.
+	vertical-align: middle;
```
Volker-E, line 108: `middle` or `top`. We've got standardized icons of 20x20px. When the icon is featuring a wrong canvas, we'd see it better with `top`?!

src/components/icon/Icon.snap.ts
AnneTee: Yeah, I can see how it would make more sense to set black as the default color since the developer can easily set the `iconColor` prop to the `currentColor` keyword or to a specific color. Updated in the latest commit.

src/components/icon/Icon.stories.ts
AnneTee: Done!

src/components/icon/Icon.stories.ts
AnneTee: I realized that I did it this way because for Structured Data projects we've been setting the `eslint/max-attributes-per-line` to 2, but that rule is actually turned off completely for WVUI. The setting we've used on SD projects has worked well, in my opinion, because it allows you to keep elements with 2 or fewer attributes to a single line, while making it _much_ easier to scan through attributes for elements that have >2. Consider the input element in our new `AutocompleteSearchInput` component:

```
<input
	:id="inputElementId"
	ref="input"
	v-model="value"
	dir="auto"
	class="wbmi-input__input"
	type="text"
	role="combobox"
	aria-autocomplete="list"
	:aria-owns="lookupResultsElementId"
	:aria-expanded="isExpanded"
	:aria-activedescendant="activeLookupItemId"
	:placeholder="placeholder"
	@input="onInput"
	@focus="onFocus"
	@blur="onBlur"
	@keyup.enter="onSubmit"
	@keyup.up="onArrowUp"
	@keyup.down="onArrowDown"
>
```

It's easier to work on, say, all of the aria attributes, or the event handlers, when you can quickly see what's going on based on the separate lines and the attribute order/groupings.

Putting all the attributes on one line when they're all static makes sense, but it would make it impossible to set a `max-attributes-per-line` limit for that project, which would (IMO) decrease template readability overall.

This seems like a good case for team-specific eslint settings. I'm happy to change the Icon template if needed, although it looks like other components are using similar template formatting.

I will admit that the hanging `>` at the end drives me crazy though 😂

src/components/icon/Icon.stories.ts
AnneTee: For the margin-bottom, but since that's browser-specific I went ahead and removed it and added an explicit margin-bottom on the div

src/components/icon/Icon.snap.ts
AnneTee: I reverted this (and width) back to 20px. I know there's more to think about here but I agree with your other comment saying that's better handled in a separate Phab ticket/PR

src/components/icon/Icon.vue
AnneTee: Done

src/components/icon/Icon.vue
AnneTee: Removed in the latest commit

src/components/icon/Icon.vue
AnneTee: 😄 I don't know enough about all the use cases to decide which is better, but perhaps this should be determined in future work on handling icon sizes?

readme.md
```diff
@@ -679,6 +689,7 @@ The expectations for submitting a patch are:
     [when running Storybook](https://github.com/storybookjs/storybook/issues/4853).
 -   If Storybook encounters an error when booting, it does not launch even after the error is
     resolved.
+-   Dead code is not type checked when building the library.
```
Volker-E, line 692: I'm confused by this comment. Isn't that self-explaining? What's the definition of 'dead code'?

src/components/icon/Icon.snap.ts
Volker-E: Yes, sounds good to me. Let's keep it simple for now.

src/components/icon/Icon.stories.ts
Volker-E: I think that's a good call, there must be some middle ground. I was thinking to have elements with Vue attributes feature line breaks, all static HTML/SVG would not.
```
    <g
      fill="currentColor"
    >
      <path
        d="path should flip"
      />
    </g>
```
This is incredibly hard to read. I feel strongly that it'll be also a source of frustration, as one exchanges easier(?) orientation in attributes with harder one in elements!

src/components/icon/Icon.stories.ts
```diff
@@ -0,0 +1,83 @@
...
+				icon: null
+			};
+		},
+		beforeMount() {
+			window.document.documentElement.setAttribute( 'dir', 'ltr' );
+		},
+		template: `
+		<div>
+			<div v-for="icon in Object.keys( icons )" :key="icon">
+				<p>
```
Volker-E: Out of curiousity: Why does it need an additional `p` here?

src/components/icon/Icon.stories.ts
AnneTee: Oh yeah, that looks bad...it might be possible to change the formatting of these generated snapshot files via jest config.

readme.md
AnneTee: See [this commit](https://github.com/wikimedia/wvui/pull/47/commits/5e2185c9f8704a2e5f0d6929e16035e3bb3041d2), I combined this line with the line below it and added a bit more info to clarify, let me know if it needs further changes!

src/components/icon/Icon.vue
Volker-E: Filed at https://phabricator.wikimedia.org/T260615

src/components/icon/Icon.stories.ts
Volker-E: To be continued at https://phabricator.wikimedia.org/T260615

src/components/icon/Icon.vue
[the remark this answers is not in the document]
Volker-E: Filed at https://phabricator.wikimedia.org/T260617

src/components/icon/Icon.vue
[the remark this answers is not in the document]
santhoshtr: We cannot do it for lang since that is not a 'style' property, but an html element attribute having its effect inherited to its child nodes.

santhoshtr (approved):
I think the current approach for dir and lang is good enough. As the library matures, I would expect addition of some utility libraries for find direction of a given language code, autonym of a given language etc. That will help to compute the dir more precisely from passed language prop. 

In somewhat distant future, [Intl](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl) will get [getDirection](https://github.com/tc39/ecma402/issues/46) and we can start using it without the need for any special libraries. Currently at MW, ULS extension exposes apis for finding autonym, scripts and language direction based on [language-data](https://github.com/wikimedia/language-data ) library.

As for developers using this library, general i18n best practices need to be followed, specifically a[dding lang and dir attributes](https://www.w3.org/International/geo/html-tech/tech-lang.html) whenever a language context switch happens.

PR #49, opened 2020-07-10
[dev] Add wikimedia-ui theme

speedandfunction-anribolon:
All incoming component patches (https://github.com/wikimedia/wvui/pull/46, https://github.com/wikimedia/wvui/pull/47) use the same less theme that imports `wikimedia-ui-base.less` and adds some variables to be used for styling components. I've decided to put it to separate patch so that it can be used widely.

### DIFF
--- a/src/themes/wikimedia-ui.less
+++ b/src/themes/wikimedia-ui.less
@@ -1,0 +1,18 @@
+@import (reference) '~wikimedia-ui-base/wikimedia-ui-base.less';
+
+@font-size-browser: 16; // assumed browser default of `16px`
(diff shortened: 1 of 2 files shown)


src/themes/wikimedia-ui.less
```diff
@@ -0,0 +1,18 @@
+@import (reference) '~wikimedia-ui-base/wikimedia-ui-base.less';
```
niedzielski, line 1: I'm not sure on the quiet colors but everything else here seems to match OOUI. We can tweak as needed later!

PR #57, opened 2020-07-17
Update/input icon

speedandfunction-anribolon:
A basic implementation of icon that can be embedded to the input component. When https://github.com/wikimedia/wvui/pull/47 is ready and merged, we will be able to use `<wvui-icon>` component.


src/components/input/Input.vue
```diff
@@ -30,8 +61,32 @@ export default Vue.extend( {
 		disabled: {
 			type: Boolean,
 			default: false
+		},
+		icon: {
+			type: String,
+			default: null
+		},
+		iconOffset: {
```
niedzielski, line 69: Do you think we'll need this? How are other libraries handling it? A quick look at [Buefy](https://github.com/buefy/buefy/blob/dev/src/components/input/Input.vue#L41) doesn't seem to need it.

src/components/input/Input.vue
```diff
@@ -44,6 +99,43 @@ export default Vue.extend( {
...
+				const $input = this.$refs.input as HTMLInputElement;
+
+				$input.focus(); // eslint-disable-line no-jquery/no-event-shorthand
+			} );
+		},
+
+		/*
+		* Adjusts input's left padding if icon is provided
+		* */
+		adjustIcon: function (): void {
```
niedzielski, line 117: Is this possible to do with CSS only?

src/components/input/Input.vue
```diff
@@ -121,4 +226,22 @@ export default Vue.extend( {
 		}
 	}
 }
+// Temp: hardcoded icon styles from mw-components
+.wvui-icon {
+	align-items: center;
+	// stylelint-disable
```
niedzielski, line 232: This disable all rules. Is it possible to just list the individual lines needed?

src/components/input/Input.vue
```diff
@@ -121,4 +226,22 @@ export default Vue.extend( {
...
+	display: inline-flex;
+	font-feature-settings: 'liga';
+	// stylelint-enable
+	font-size: 1.5em;
+	justify-content: center;
+	letter-spacing: normal;
+	line-height: 1;
+	text-indent: 0;
+	-webkit-transition: 0.3s cubic-bezier(0.25, 0.8, 0.5, 1), visibility 0s;
```
niedzielski, line 241: Doesn't autoprefixer do this stuff?

src/components/input/Input.vue
```diff
@@ -57,6 +149,15 @@ export default Vue.extend( {
 	vertical-align: middle;
 	box-sizing: border-box;
 
+	&__icon {
+		position: absolute;
+		padding-left: 9px;
+		top: 50%;
+		transform: translateY(-50%);
+		line-height: 1;
+		text-align: center;
```
Volker-E, line 158: Why is a `text-align` needed in an icon element?

src/components/input/Input.vue
Volker-E: Also, where does the non-standard transition originate from?

speedandfunction-anribolon: @Volker-E  @niedzielski I've just reused a snippet from [this](https://github.com/niedzielski/mw-components/blob/2dcc3a1510a3d665de362b8cbdbe49cac7ab3957/src/components/icon/Icon.vue#L64) prototype (compiled version). I put it here as a placeholder which will be replaced with wvui-icon component. My intention was not to put a lot of effort into it, so I've just copy-paste it from [storybook](https://mw-components.netlify.app/ui/?path=/story/components-icon--all-icons)

src/components/input/Input.vue
speedandfunction-anribolon: According to previous designs provided by Alex, I though we might need it. But now I believe we can remove it.

.storybook/preview.less
```diff
@@ -5,3 +5,7 @@ body {
 	font-size: 0.875em;
 	color: @wmui-color-base10;
 }
+
+.sb-input-preview {
```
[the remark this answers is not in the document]
speedandfunction-anribolon: Good idea, I agree

src/components/input/Input.vue
niedzielski: @speedandfunction-anri, please put effort into it :smile: We're trying to achieve a reasonable baseline that adheres to at least the OOUI styles needed for the search form. Please see these notes for the step-by-step:

https://phabricator.wikimedia.org/T255902
https://www.mediawiki.org/wiki/Vue.js/OOUI_migration_guide

src/components/input/Input.vue
speedandfunction-anribolon: OOUI uses something similar to calculate padding if label is provided. I'm still not sure if we'll need label for inputs. What do you think?

src/components/input/Input.vue
speedandfunction-anribolon: It's not, you're right

.storybook/preview.less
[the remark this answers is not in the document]
niedzielski: Sorry, it wasn't clear if this style should be moving then or not.

src/components/input/Input.vue
niedzielski: :+1:

src/components/input/Input.vue
niedzielski: @Volker-E, do we need this configurability? Vuefy seems to take the approach of just start (left) or end (right): https://github.com/buefy/buefy/blob/dev/src/components/input/Input.vue#L155

.storybook/preview.less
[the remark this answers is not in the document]
speedandfunction-anribolon: Moved to Input.stories.less

src/components/input/Input.vue
speedandfunction-anribolon: @niedzielski I've replaced that styles with styles from Anne's icon PR.

src/components/input/Input.vue
speedandfunction-anribolon: It's not relevant anymore, I've replaced this with CSS

src/components/input/Input.vue
```diff
@@ -44,6 +97,16 @@ export default Vue.extend( {
 		},
 		onBlur( event: FocusEvent ): void {
 			this.$emit( 'blur', event );
+		},
+		/*
+		* Sets focus to input if icon is clicked
+		* */
+		onIconClick(): void {
```
niedzielski, line 104: I think this could be replaced by setting `pointer-events: none;` on the __icon.

speedandfunction-anribolon: Yes, nice catch, didn't know it would work!

PR #58, opened 2020-07-17
[update] Add button control for input

speedandfunction-anribolon:
Add support of using `<wvui-button>` as a slotted component for `<wvui-input>`.


src/components/input/Input.snap.ts
```diff
@@ -25,6 +27,58 @@ exports[`matches the snapshot Case 1 Input Type: search : ({"type": "search"}) =
...
+   
+  <span
+    class="wvui-input__icon"
+  >
+    <span
+      class="wvui-icon"
+    >
+      <svg
+        fill="#72777d"
+        height="24"
```
Volker-E, line 53: Icons are 20x20 everywhere!

src/components/input/Input.snap.ts
```diff
@@ -25,6 +27,58 @@ exports[`matches the snapshot Case 1 Input Type: search : ({"type": "search"}) =
...
    
   <!---->
+   
+  <!---->
+</div>
+`;
+
+exports[`should render a button in a slot 1`] = `
+<div
+  class="wvui-input wvui-input--has-control"
```
Volker-E, line 37: That's more a general comment in contrast to the icon issue below:
I find this absolutely horrible to read, devs are writing and debugging HTML for years, I don't think we should outsmart common writing styles here. :/

src/components/input/Input.vue
```diff
@@ -225,6 +245,29 @@ export default Vue.extend( {
...
+			border-top-left-radius: 0;
+			border-bottom-left-radius: 0;
+			border-left-width: 0;
+		}
+	}
+
+	&--has-control {
+		// stylelint-disable-next-line plugin/no-unsupported-browser-features
+		display: flex;
+		flex-direction: row;
```
niedzielski, line 263: I believe row is the initial value.

src/components/input/Input.vue
```diff
@@ -225,6 +245,29 @@ export default Vue.extend( {
 			border-color: @border-color-input--hover;
 		}
 	}
+
+	&__control {
+		height: @size-base;
+
+		& button {
```
niedzielski, line 252: Please consider adding some documentation for these new styles. They make sense in the context of this distinct pull request but may be harder to interpret later. My read is that these are for eliminating the gap between the button and the input.

src/components/input/Input.snap.ts
speedandfunction-anribolon: @Volker-E I was referencing to https://github.com/buefy/buefy/blob/dev/src/components/input/Input.vue#L155. Which naming would you suggest?

src/components/input/Input.snap.ts
speedandfunction-anribolon: Fixed

src/components/input/Input.snap.ts
niedzielski: Sorry but is this part of a larger conversation I missed? What is the subject here? `wvui-input--has-control`, multiple classes, or something else?

Volker-E: Writing HTML tags not in one go, having line-breaks in there. It's so unfamiliar :(

niedzielski: I see. This code is automatically generated from Jest. I was going to say that I'm not sure what options we have to beautify it, if any, as it's not even plain HTML but rather an HTML string. However, these projects seem relevant and make me think that something is possible:

- https://github.com/rayrutjes/jest-serializer-html
- https://github.com/tanem/jest-prettyhtml-matchers

I don't think we should explore these right now, unfortunately, because I told Daniel we'd try not to change the NPM dependencies between now and release (he's working with Security). However, if you make a PR, maybe we can have it "on deck" for ready for merging post-deployment.

niedzielski: I'm marking this conversation resolved to keep this PR moving forward. @Volker-E, please feel free to open a PR or task to track the above improvements.

src/components/input/Input.vue
```diff
@@ -78,8 +85,15 @@ export default Vue.extend( {
 
 	},
 	computed: {
-		rootClasses(): Record<string, boolean> {
+		/*
+		* Checks if slot component is provided
+		* */
+		hasControlSlot: function (): boolean {
```
niedzielski, line 91: Very minor: Any reason to use functions instead of method here?

src/components/input/Input.vue
```diff
@@ -78,8 +85,15 @@ export default Vue.extend( {
...
+			return !!this.$scopedSlots.default;
+		},
+		rootClasses: function () {
```
niedzielski, line 94: Please consider adding return types here.

src/components/input/Input.vue
```diff
@@ -189,6 +203,29 @@ export default Vue.extend( {
...
+
+	&__button {
+		height: @size-base;
+
+		// Eliminate the gap between the button and the input.
+		& .wvui-button {
+			height: 100%;
+			border-top-left-radius: 0;
+			border-bottom-left-radius: 0;
+			border-left-width: 0;
```
nicholasray, line 215: This style ends up getting overridden  by `.wvui-button:not( [ disabled ] ).wvui-button--framed`. Is that intended?

nicholasray: Sorry I didn't mean to highlight all of those lines, just  specifically L215

speedandfunction-anribolon: @nicholasray I've added `wvui-button--framed` class to the selector.

src/components/input/Input.vue
```diff
@@ -78,8 +85,15 @@ export default Vue.extend( {
 
 	},
 	computed: {
+		/*
+		* Checks if slot component is provided
+		* */
```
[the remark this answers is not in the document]
speedandfunction-anribolon: Fixed

niedzielski:
@speedandfunction-anribolon, I've changed the destination branch from update/input-icon, which I just merged, to master but now there is a merge conflict.

src/components/input/Input.stories.ts
```diff
@@ -113,7 +113,7 @@ export const withClearAction = (): Vue.Component =>
 		template: `
 		<div class="sb-input-preview">
 			<wvui-input
-				placeholder="Search..."
+				placeholder="Type something..."
```
Volker-E, line 116: Minor rec: `...` should be `…`, a typographically correct ellipsis. One is three chars, latter one.

src/components/input/Input.vue
```diff
@@ -101,10 +110,14 @@ export default Vue.extend( {
 			type: String,
 			default: null
 		},
-		// clearable properyy will override indicator property
+		// clearable property will override indicator property
```
Volker-E, line 113: `.` at eol

src/components/input/Input.vue
```diff
@@ -117,15 +130,58 @@ export default Vue.extend( {
...
 				'wvui-input--has-icon': !!this.icon,
-				'wvui-input--clearable': this.clearable
+				'wvui-input--clearable': this.clearable,
+				'wvui-input--has-button': !!this.$scopedSlots.default
 			};
 		},
-		iconName(): string {
-			return this.clearable ? 'clear' : this.icon;
+
+		/*
```
niedzielski, line 157: Is this only needed to support slotted content that changes (the button in this case)? Is there another approach we should take to avoid supporting this use case? If not, did you consider pulling this out into a directive? [Directive example from Vuetify](https://github.com/vuetifyjs/vuetify/blob/78b0ae2bd2d8b843417250f0cd3bcecf99604112/packages/vuetify/src/directives/mutate/index.ts).

This seems a little unusual to me so I wanted to do a brief survey. I see that no MutationObservers are present in [ContentTranslation (super multilingual app)](https://github.com/wikimedia/mediawiki-extensions-ContentTranslation/search?q=MutationObserver&unscoped_q=MutationObserver), [Buefy](https://github.com/buefy/buefy/search?q=MutationObserver&unscoped_q=MutationObserver), [KeenUI](https://github.com/JosephusPaye/Keen-UI/search?q=MutationObserver&unscoped_q=MutationObserver), [vux](https://github.com/airyland/vux/search?q=MutationObserver&unscoped_q=MutationObserver), [fish-ui](https://github.com/myliang/fish-ui/search?q=MutationObserver&unscoped_q=MutationObserver), and [muse-ui](https://github.com/museui/muse-ui/search?q=MutationObserver&unscoped_q=MutationObserver). MutationObservers _are_ used in [a number of places in vue-material](https://github.com/vuematerial/vue-material/search?q=MdObserveElement&unscoped_q=MdObserveElement), [infinite scroll in ElemeFE](https://github.com/ElemeFE/element/search?q=MutationObserver&unscoped_q=MutationObserver). [tabs in ViewUI](https://github.com/view-design/ViewUI/search?q=MutationObserver&unscoped_q=MutationObserver), and [as a utility in several places bootstrap-vue](https://github.com/bootstrap-vue/bootstrap-vue/search?q=observe-dom&unscoped_q=observe-dom).

I think we can do whatever we need to here but please share the options you're thinking about and document them. I wanted to make sure this was necessary as my gut feeling was that it was a little heavy-duty compared to the usual declarative nature of assembling Vue components in a static configuration.

src/components/input/Input.vue
```diff
@@ -258,6 +356,25 @@ export default Vue.extend( {
 		}
 	}
 
+	&--has-button {
```
Volker-E, line 359: So we're using these as indicators for combined components I __guess__. This should be clarified beyond current patch: https://phabricator.wikimedia.org/T259203

src/components/input/Input.vue
```diff
@@ -220,7 +316,9 @@ export default Vue.extend( {
 			border-color: @border-color-base--disabled;
 
 			& ~ .wvui-input__icon,
-			& ~ .wvui-input__indicator {
+			& ~ .wvui-input__indicator,
+			& ~ .wvui-input__button {
+				pointer-events: none;
```
Volker-E, line 321: In OOUI we've followed best-practice to limit properties only to directly intended selectors. In my understanding that would mean here: `& ~ .wvui-input__button` only…?

src/components/input/Input.stories.ts
Volker-E: Same applies to Search... above

niedzielski:
@speedandfunction-anribolon, do you have what you need for this patch? It looks like there's still open feedback.

src/components/input/Input.vue
```diff
@@ -39,12 +39,14 @@
...
 		</span>
 		<span
-			v-if="(clearable && currentValue && !disabled) || indicator"
 			ref="indicator"
 			class="wvui-input__indicator"
 			@click="onClear"
 		>
-			<span class="wvui-icon">
+			<span
+				v-if="hasIndicator || isClearable"
```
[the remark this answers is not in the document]
speedandfunction-anribolon: The main idea is that we don't need to show `clear` indicator if there is no value in the input. This `return !this.clearable && !!this.indicator;` basically means that `clearable` prop will override `indicator` prop. Does this make sense @niedzielski ? I can add some docs.

src/components/input/Input.stories.ts
```diff
@@ -123,3 +123,56 @@ export const withClearAction = (): Vue.Component =>
...
+				placeholder="Search..."
+				:disabled="disabled"
+			>
+				<wvui-button>Search</wvui-button>
+			</wvui-input>
+		</div>
+	`
+	} );
+
+const searchLanguageMap = {
```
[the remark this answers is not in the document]
speedandfunction-anribolon: I'm working on a separate PR which will add RTL support to the whole input and all related elements. I will add Hebrew there @niedzielski

src/components/input/Input.vue
```diff
@@ -59,6 +61,13 @@
 				</svg>
 			</span>
 		</span>
+		<span
+			v-if="hasControlSlot"
+			ref="button"
+			class="wvui-input__button"
```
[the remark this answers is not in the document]
speedandfunction-anribolon: I will fix it, you're right

src/components/input/Input.vue
speedandfunction-anribolon: The main idea of using MutationObserver was to observe width of the slotted button, so that if you change a language, button width is also changed and the component adjusts `clear` button, otherwise search button overlaps `clear` button. Another option is to call `adjustIndicator` in `updated` lifecycle hook as it react on all changes for all slots. But my consern is that `updated` hook is called every time user types in the input. Actually I like the idea of moving MutationObserver to a directive. @niedzielski

src/components/input/Input.stories.ts
speedandfunction-anribolon: Fixed @Volker-E

src/components/input/Input.snap.ts
```diff
@@ -12,6 +12,12 @@ exports[`matches the snapshot Case 0 Input Type: text : ({"type": "text"}) => HT
    
   <!---->
    
+  <span
+    class="wvui-input__indicator"
```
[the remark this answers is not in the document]
speedandfunction-anribolon: Fixed

src/components/input/Input.vue
```diff
@@ -101,10 +110,14 @@ export default Vue.extend( {
...
 		},
-		// clearable properyy will override indicator property
+		// clearable property will override indicator property
 		clearable: {
 			type: Boolean,
 			default: false
+		},
+		buttonLabel: {
+			type: String,
+			default: null
```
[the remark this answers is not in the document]
speedandfunction-anribolon: I've removed this prop, it wasn't necessary at all

src/components/input/Input.vue
speedandfunction-anribolon: @Volker-E Actually I want this properties to be applied to both `wvui-input__icon` and `wvui-input__indicator` if the input (which is sibling selector for both) is disabled. I've removed `wvui-input__button`

src/components/input/Input.stories.ts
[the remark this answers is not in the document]
speedandfunction-anribolon: Done

src/components/input/Input.vue
niedzielski: Can you take a look at [this alternative approach](https://github.com/wikimedia/wvui/pull/75/files)? I think having the slot within the input component is making this more complex than it needs to be for many use cases. So far as I can tell, a consumer needs to only supply a wrapper with `display: flex` to use a `wvui-input` and `wvui-button` on the same line. I think this really simplifies a lot of the code but maybe I'm missing something. Please let me know your thoughts.

src/components/input/Input.vue
Volker-E: As it's filed as task, no further actions on this patch for now.

src/components/input/Input.vue
```diff
@@ -117,22 +126,76 @@ export default Vue.extend( {
...
+				'wvui-input--clearable': this.clearable,
+				'wvui-input--has-button': !!this.$scopedSlots.button
 			};
 		},
-		iconName(): string {
-			return this.clearable ? 'clear' : this.icon;
+
+		/*
+		* Creates a MutationObserver instance for adjusting indicator position
+		* according to button slot DOM changes
```
Volker-E, line 155: Comments are inconsistently featuring a `.` at the end. It should be every comment eol

src/components/input/Input.vue
speedandfunction-anribolon: @niedzielski I've pushed all changes related to this approach, you're right, I over-complicated implementation using slot.

src/components/input/Input.vue
```diff
@@ -39,12 +39,14 @@
 			</span>
 		</span>
 		<span
-			v-if="(clearable && currentValue && !disabled) || indicator"
+			v-if="isClearable || hasIndicator"
 			ref="indicator"
 			class="wvui-input__indicator"
 			@click="onClear"
```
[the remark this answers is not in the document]
speedandfunction-anribolon: changed

src/components/input/Input.vue
speedandfunction-anribolon: I've removed that comment.

src/components/input/Input.stories.ts
```diff
@@ -123,3 +123,54 @@ export const withClearAction = (): Vue.Component =>
...
+			<wvui-button :disabled="disabled">Search</wvui-button>
+		</div>
+	`
+	} );
+
+const searchLanguageMap = {
+	English: 'Search',
+	Russian: 'Искать',
+	Vietnamese: 'Tìm kiếm',
+	Japaneese: '探す',
```
niedzielski, line 148: Little typo. This should be "Japanese".

speedandfunction-anribolon: You're right. Fixed!

src/components/input/Input.snap.ts
```diff
@@ -32,87 +32,87 @@ exports[`matches the snapshot Case 1 Input Type: search : ({"type": "search"}) =
...
     class="wvui-input__input"
     dir="auto"
     type="text"
   />
    
-  <!---->
-   
   <span
-    class="wvui-input__indicator"
+    class="wvui-input__start-icon"
```
Volker-E, line 46: Why does the name contain 'start'?

src/components/input/Input.snap.ts
```diff
@@ -32,87 +32,87 @@ exports[`matches the snapshot Case 1 Input Type: search : ({"type": "search"}) =
...
   >
     <span
       class="wvui-icon"
     >
       <svg
         aria-hidden="true"
-        height="12"
+        height="20"
         role="presentation"
```
Volker-E, line 54: I know, we'll integrate this shortly from the icon component, but `role` attribute is superfluos if `aria-hidden`. Decide for yourself if you want to move it forward without changing now…

src/components/input/Input.snap.ts
Volker-E: Would icon-start and icon-end also be simpler readable as attributes?

niedzielski: @Volker-E, sorry, I'm having trouble following this thread as fellow reviewer so bear with me. I think I've somehow managed to misunderstand every single one of points! :coffee:

As I understand it, there are four topics here:

1. The input's wvui-icon sub-component class naming.
2. The input's wvui-icon classes and code sharing.
3. Input's composition of wvui-icon components vs pseudo-element usage.
4. Passing icons as input attributes.

1: When you talk about the naming, do you mean the choice to use "start-icon" itself as opposed to say "before-icon"? OR, maybe you mean the hyphenation style? Or maybe even indicator vs icon terminology? Something else? I would guess that any name you came up with would be a welcome follow-up.

2: Regarding wvui-icon classes here, is my understanding correct that you're proposing:

```less
.wvui-input {
  .wvui-icon {
    &--start {
        background: red;
    }
    &--end {
        background: green;
    }
  }
}
```

3: On the topic of `:before` and `:after` structure, I agree this would be ideal from an HTML point-of-view. I personally haven't had a chance to explore pseudo-element composition though which I think would be my most pressing concern. If this subject is important to you, I recommend opening a new spike to 1) survey pseudo-element usage in some popular Vue.js libraries, 2) research Vue.js's pseudo-element capabilities, 3) document the approximate number of icon compositions in the OOUI library or OOUI-based projects.

4: Finally, regarding attributes, wvui-input accepts the start and end icons as attributes. Is something more wanted?

speedandfunction-anribolon: Not sure if using pseudo-elements is a good idea. Few concerns and questions:
1. How to pass icons to <wvui-input> component?
2. We can't apply any event listeners to pseudo-elements.

src/components/input/Input.snap.ts
speedandfunction-anribolon: @Volker-E I've removed all stubs.

src/components/input/Input.snap.ts
```diff
@@ -45,19 +45,20 @@ exports[`should render a clear icon 1`] = `
...
+        height="20"
         viewBox="0 0 20 20"
-        width="12"
+        width="20"
         xmlns="http://www.w3.org/2000/svg"
       >
+        <!---->
+         
         <g
           fill="#72777d"
```
Volker-E, line 63: The fills should be `#000`. We're manipulating opacity on interaction.

PR #60, opened 2020-07-21
Remove `border-radius-base` override

Volker-E:
`border-radius-base` is `2px` and part of WikimediaUI Base.


niedzielski:
Nice one-liner. Looks like the import structure you proposed can be seen working very well here :+1:

PR #63, opened 2020-07-22
[component][input] Add indicator and clear action

speedandfunction-anribolon:
Add indicator (right icon) to input component. It's considered to be `<wvui-icon>` component.
Add `clearable` property which overrides `indicator` prop and adds an icon which clears input value.


src/components/input/Input.vue
```diff
@@ -107,6 +154,19 @@ export default Vue.extend( {
...
+		},
+		onClear(): void {
+			if ( !this.clearable ) {
+				return;
+			}
+			this.$emit( 'input', '' );
+			this.setCurrentValue( '' );
+		},
+		setCurrentValue( value: string | number ): void {
+			if ( value === this.currentValue ) {
```
niedzielski, line 166: Can you comment explaining why this check is needed?

src/components/input/Input.vue
```diff
@@ -107,6 +154,19 @@ export default Vue.extend( {
 
 				$input.focus(); // eslint-disable-line no-jquery/no-event-shorthand
 			} );
+		},
+		onClear(): void {
+			if ( !this.clearable ) {
```
niedzielski, line 159: Do we need this check? Why would this would be invoked if the state isn't already clearable?

src/components/input/Input.vue
```diff
@@ -120,14 +180,30 @@ export default Vue.extend( {
...
+	&__icon {
+		padding-left: @padding-horizontal-input-text;
+	}
+
+	&__indicator {
+		right: 0;
+		padding-right: @padding-horizontal-input-text;
+	}
+
+	&--clear {
```
niedzielski, line 201: I think this is a more generic state for whenever the indicator is clickable and a hand icon should be shown. Is there another term we should use instead of `--clear`?

src/components/input/Input.vue
speedandfunction-anribolon: If input is not clearable (just has an indicator) we don't want to clear value.

src/components/input/Input.vue
speedandfunction-anribolon: It's not needed actually

niedzielski: Remove then?

speedandfunction-anribolon: Removed

src/components/input/Input.vue
speedandfunction-anribolon: changed to `clearable`

PR #68, opened 2020-07-27
Component/search suggestion

speedandfunction-anribolon:
This patch introduces <wvui-typeahead-suggestion> component, which is supposed to be used to represent a single search result for Typeahead Search. For inspiration I've been using [this prototype](https://di-searchland.web.app/) and Portals repo. 

Bug: [T256042](https://phabricator.wikimedia.org/T256042)


.eslintignore
```diff
@@ -22,3 +22,6 @@
 
 # Ignore Jest snapshots.
 *.snap.ts
+
+# Ignore json files with mock data
+/src/components/typeahead-suggestion/TypeaheadSuggestion.stories.json
```
niedzielski, line 27: What is being ignored here?

src/components/typeahead-suggestion/TypeaheadSuggestion.stories.less
```diff
@@ -0,0 +1,24 @@
+.sb-suggestions-list {
```
niedzielski, line 1: Is it intentional to style this differently when outside of .sb-search-wrapper? I see more rules below but wasn't clear on the intent. Please add a couple docs.

src/components/typeahead-suggestion/TypeaheadSuggestion.stories.ts
```diff
@@ -0,0 +1,99 @@
...
+			suggestion(): TypeaheadSuggestion {
+				const suggestion = suggestionsList.pages[ 1 ];
+				return {
+					...suggestionsList.pages[ 1 ],
+					thumbnail: this.thumbnail ? suggestion.thumbnail : null
+				};
+			}
+		},
+		template: `
+		<ul class="sb-suggestions-list" role="listbox">
```
niedzielski, line 35: I think this should be an ordered list as the results are ordered by relevance.

src/components/typeahead-suggestion/TypeaheadSuggestion.vue
```diff
@@ -0,0 +1,172 @@
+<template>
+	<li
+		v-if="suggestion"
+		class="wvui-typeahead-suggestion"
+		:class="rootClasses"
+		v-bind="$attrs"
```
niedzielski, line 6: Won't this get the attributes as the outermost component?

src/components/typeahead-suggestion/TypeaheadSuggestion.vue
```diff
@@ -0,0 +1,172 @@
...
+	>
+		<a :href="suggestionWikiLink" class="wvui-typeahead-suggestion__link">
```
niedzielski, line 8: Should this be the parent instead of the list item?

.eslintignore
speedandfunction-anribolon: Thumbnail urls are pretty long so I decided to ignore the whole file. Do you think we need to care of formatting for mock data files?

src/utils/Utils.ts
```diff
@@ -0,0 +1,26 @@
+export default class WvuiUtils {
```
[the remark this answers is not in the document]
speedandfunction-anribolon: @niedzielski  We have ESLint rule `ES2015 module syntax is preferred over custom TypeScript modules and namespaces.(@typescript-eslint/no-namespace)`

.eslintignore
niedzielski: I don't mind ignoring it if there's a good reason but right now the comment documents the response not the motivation.

src/components/typeahead-suggestion/TypeaheadSuggestion.snap.ts
```diff
@@ -0,0 +1,101 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`matches the snapshot Case 0 With thumbnail: ({"suggestion": [Object]}) => HTML 1`] = `
+<li
```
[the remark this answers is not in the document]
speedandfunction-anribolon: I've replaced <li> with <a>, so that anchor tag is the root element

src/components/typeahead-suggestion/TypeaheadSuggestion.snap.ts
```diff
@@ -0,0 +1,101 @@
...
+  class="wvui-typeahead-suggestion"
+>
+  <a
+    class="wvui-typeahead-suggestion__link"
+    href="/wiki/Obsessive–compulsive_disorder"
```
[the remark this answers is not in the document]
speedandfunction-anribolon: Done

src/components/typeahead-suggestion/TypeaheadSuggestion.stories.less
```diff
@@ -0,0 +1,24 @@
...
+	max-width: 500px;
+	list-style: none;
+	margin: 0;
+	background: #fff;
+	border: 1px solid #a2a9b1;
+	border-radius: 0 0 2px 2px;
+	padding: 0;
+}
+
+.sb-search-wrapper {
```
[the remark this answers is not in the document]
speedandfunction-anribolon: Done

.eslintignore
speedandfunction-anribolon: I've removed it from ignoring in the end

src/components/typeahead-suggestion/TypeaheadSuggestion.stories.ts
speedandfunction-anribolon: Replaced

src/components/typeahead-suggestion/TypeaheadSuggestion.vue
speedandfunction-anribolon: Yeah, it was unnecessary.

speedandfunction-anribolon: I agree, replaced

src/components/typeahead-suggestion/TypeaheadSuggestion.vue
speedandfunction-anribolon: You're right it is unnecessary.

src/components/typeahead-suggestion/TypeaheadSuggestion.stories.less
speedandfunction-anribolon: Done

src/utils/Utils.ts
```diff
@@ -0,0 +1,26 @@
...
+				return '&lt;';
+			case '>':
+				return '&gt;';
+			case '&':
+				return '&amp;';
+			default:
+				return value;
+		}
+	}
+}
```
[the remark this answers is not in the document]
speedandfunction-anribolon: Done

src/components/typeahead-suggestion/TypeaheadSuggestion.ts
```diff
@@ -0,0 +1,14 @@
+export interface TypeaheadSuggestion {
+	key: string;
+	description: string | null;
```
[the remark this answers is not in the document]
speedandfunction-anribolon: You're right

src/components/typeahead-suggestion/TypeaheadSuggestion.stories.less
```diff
@@ -0,0 +1,24 @@
...
+	position: relative;
+	width: 500px;
+
+	.sb-suggestions-list {
```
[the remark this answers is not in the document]
speedandfunction-anribolon: I've changed class names.

src/components/typeahead-suggestion/TypeaheadSuggestion.vue
```diff
@@ -0,0 +1,172 @@
...
+			<span
```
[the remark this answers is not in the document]
speedandfunction-anribolon: @niedzielski Actually I use the same approach as [portals](https://wikipedia.org). Honestly saying I'm not sure as for all details, but as far as i remember there are some performance differences in using background-image. @Volker-E can you comment please?

src/utils/StringUtils.ts
```diff
@@ -0,0 +1,36 @@
+export default class WvuiStringUtils {
```
niedzielski, line 1: Can we omit this class and export each function instead? A class adds an extra level of indentation here and will require a private constructor. If you like a single export, maybe we can use a wildcard import instead like `import * as StringUtil from './StringUtil'`.

src/components/typeahead-suggestion/TypeaheadSuggestion.ts
```diff
@@ -0,0 +1,14 @@
+export interface TypeaheadSuggestion {
```
niedzielski, line 1: Please document these types and their properties. For example, I'm not sure what the difference between width and size are. I think you've just taken the results from the API and typed them. That totally makes sense :+1: but I think we can safely assume that we'll have an "ideal type" when passing data to a component so you should make those types instead.

src/components/typeahead-suggestion/TypeaheadSuggestion.vue
```diff
@@ -0,0 +1,133 @@
...
+	props: {
+		active: {
+			type: Boolean,
+			default: false
+		},
+		query: {
+			type: String as PropType<string | undefined>,
+			default: undefined
+		},
+		suggestion: {
```
niedzielski, line 48: Do you think suggestion should be required?

src/components/typeahead-suggestion/TypeaheadSuggestion.vue
```diff
@@ -0,0 +1,133 @@
...
+	name: 'WvuiTypeaheadSuggestion',
+	components: { WvuiTypeaheadSuggestionTitle },
+	props: {
+		active: {
+			type: Boolean,
+			default: false
+		},
+		query: {
+			type: String as PropType<string | undefined>,
+			default: undefined
```
niedzielski, line 46: Did you consider setting the default to an empty string so this type can only be a `string` and not a `string | undefined`?

src/components/typeahead-suggestion/TypeaheadSuggestion.vue
```diff
@@ -0,0 +1,133 @@
...
+			return `/wiki/${this.suggestion?.key}`;
+		},
+		/*
+		* Generates a proper value for background-image
+		* */
+		thumbnailBackgroundImage(): string {
+			return `url(${this.suggestion?.thumbnail?.url})`;
+		}
+	},
+	methods: {
```
niedzielski, line 72: This appears to be empty now. Can we remove it?

src/components/typeahead-suggestion-title/TypeaheadSuggestionTitle.vue
```diff
@@ -0,0 +1,71 @@
...
+			default: undefined
+		}
+	},
+	computed: {
+		/*
+		* Formats title adding highlighted query if it matches
+		* */
+		titleChunks(): string[] {
+			if ( !this.query ) {
+				return [ this.title as string ];
```
niedzielski, line 32: This cast appears to be incorrect. `titleChunks()` says it only returns a string array but `title` can be undefined.

src/components/typeahead-suggestion-title/TypeaheadSuggestionTitle.vue
```diff
@@ -0,0 +1,71 @@
...
+	<!--eslint-disable-next-line vue/no-v-html-->
+	<span class="wvui-typeahead-suggestion__title">
+		<!--eslint-disable-next-line max-len-->
+		{{ titleChunks[ 0 ] }}<em v-if="titleChunks[ 1 ]" class="wvui-typeahead-suggestion__match">{{ titleChunks[ 1 ] }}</em>{{ titleChunks[ 2 ] }}
+	</span>
+</template>
+
+<script lang="ts">
+import Vue, { PropType } from 'vue';
+import WvuiUtils from '../../utils/StringUtils';
```
niedzielski, line 11: This should be StringUtils.

src/components/typeahead-suggestion-title/TypeaheadSuggestionTitle.vue
```diff
@@ -0,0 +1,71 @@
...
+		title: {
+			type: String as PropType<string | undefined>,
+			default: undefined
+		}
+	},
+	computed: {
+		/*
+		* Formats title adding highlighted query if it matches
+		* */
+		titleChunks(): string[] {
```
niedzielski, line 30: Minor: did you consider moving this logic to a new file in the component folder instead of putting it in the the component itself? The calculation seems kind of independent of the user interface and may be easier to test if it was just an isolated function.

src/components/typeahead-suggestion/TypeaheadSuggestion.vue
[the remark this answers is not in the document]
j4n-co: I can't speak to the performance impact of background-image vs img tab, but the reason it was done that way on the portals was to fit the different sized thumbnails into a square shape using `background-size: cover;`. 

If the thumbnails provided by the API will be consistently square, then we can reconsider this approach. If they'll vary in size, then I think this is still the easiest way to achieve that fitting.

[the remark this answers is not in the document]
niedzielski: They [seem](https://en.wikipedia.org/w/rest.php/v1/search/title?q=Dog) to be rectangular (capped at 200px in both dimensions), not square.

src/components/typeahead-suggestion/TypeaheadSuggestion.vue
```diff
@@ -0,0 +1,133 @@
...
+	&__thumbnail-placeholder,
+	&__thumbnail {
+		display: block;
+		min-width: @typeahead-suggestion-thumb-width;
+		min-height: @typeahead-suggestion-thumb-height;
+		max-width: @typeahead-suggestion-thumb-width;
+		max-height: @typeahead-suggestion-thumb-height;
+		border: @border-width-base @border-style-base @wmui-color-base80;
+		border-radius: @border-radius-base;
+		margin: 0 12px 0 0;
```
j4n-co, line 103: Rather than using a 12px margin here, what do you think about using `text-indent: 12px` on the `&__text` element instead? That would help i18n by proving the correct margin for RTL languages.

src/components/typeahead-suggestion/TypeaheadSuggestion.vue
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	&__thumbnail-placeholder,
+	&__thumbnail {
+		display: block;
+		min-width: @typeahead-suggestion-thumb-width;
+		min-height: @typeahead-suggestion-thumb-height;
+		max-width: @typeahead-suggestion-thumb-width;
+		max-height: @typeahead-suggestion-thumb-height;
+		border: @border-width-base @border-style-base @wmui-color-base80;
```
j4n-co, line 101: Instead of using a border on the images, let's get fancy and use a 1px drop-shadow instead. Borders tend to cut into the border-radius and it makes the border-radius look smaller on the inside of the box than the outside. Using a box-shadow disguised as a border prevents that from happening. Here's a demo of what I mean: https://codepen.io/j4n/pen/poyjMod

src/components/typeahead-suggestion/TypeaheadSuggestion.vue
```diff
@@ -0,0 +1,133 @@
...
+
+		.wvui-typeahead-suggestion__title,
+		.wvui-typeahead-suggestion__description {
+			display: block;
+		}
+
+		.wvui-typeahead-suggestion__description {
+			margin: 0;
+			color: @wmui-color-base30;
+			font-size: 13px;
```
j4n-co, line 126: Let's stick with the Mediawiki convention and define the font-sizes in variables with `em`'s like in Vector: 
e.g. `@font-size-search-description: unit( 13 / @font-size-browser, @wvui-unit ); // Equals 0.8125em.`

src/components/typeahead-suggestion-title/TypeaheadSuggestionTitle.vue
```diff
@@ -0,0 +1,71 @@
...
+} );
+</script>
+
+<style lang="less">
+@import ( reference ) '@/themes/wikimedia-ui.less';
+
+.wvui-typeahead-suggestion__title {
+	margin: 0 0 2px 0;
+	color: @wmui-color-base10;
+	font-size: 15px;
```
j4n-co, line 61: Same here with the font-size, an em-based variable would be preferable.
`@font-size-search-title: unit( 15 / @font-size-browser, @wvui-unit );`

src/components/typeahead-suggestion/TypeaheadSuggestion.stories.ts
```diff
@@ -0,0 +1,114 @@
...
+			};
+		},
+		computed: {
+			suggestionsList(): TypeaheadSuggestion[] {
+				return this.isVisible ? ( suggestionsList.pages as [] ).slice( 0, 6 ) : [];
+			}
+		},
+		methods: {
+			onInput( event: InputEvent ): void {
+				const { target } = event;
```
j4n-co, line 84: this method doesn't pass the event object, instead the parameter seems to be the value of the input. Perhaps that's because of `this.$emit( 'input', value );` in `Input.vue`?

src/utils/StringUtils.ts
speedandfunction-anribolon: Make sense to me, fixed.

src/components/typeahead-suggestion/TypeaheadSuggestion.vue
speedandfunction-anribolon: Sure, fixed

src/components/typeahead-suggestion/TypeaheadSuggestion.vue
speedandfunction-anribolon: Fixed

src/components/typeahead-suggestion-title/TypeaheadSuggestionTitle.vue
speedandfunction-anribolon: Make sense, moved to a separate file

src/components/typeahead-suggestion/TypeaheadSuggestion.vue
speedandfunction-anribolon: Fixed

src/components/typeahead-suggestion-title/TypeaheadSuggestionTitle.vue
speedandfunction-anribolon: Done

src/components/typeahead-suggestion/TypeaheadSuggestion.stories.ts
speedandfunction-anribolon: I've fixed it

src/components/typeahead-suggestion/TypeaheadSuggestion.vue
speedandfunction-anribolon: Nove catch! Done.

src/components/typeahead-suggestion-title/TypeaheadSuggestionTitle.vue
speedandfunction-anribolon: Fixed

src/components/typeahead-suggestion/TypeaheadSuggestion.ts
speedandfunction-anribolon: Done

src/components/typeahead-suggestion/TypeaheadSuggestion.ts
```diff
@@ -0,0 +1,19 @@
...
```
niedzielski, line 1: Do we need this file at all now that Bernd's patch is merged? These types appear to match those in SearchClient.

src/components/typeahead-suggestion/TypeaheadSuggestion.stories.less
```diff
@@ -0,0 +1,26 @@
+// Styles for a list of suggestions
+.sb-search__suggestions {
```
niedzielski, line 2: Any reason not to nest this within `.sb-search` like `&__suggestions`? I think we've been doing that for the components themselves.

src/components/typeahead-suggestion-title/TypeaheadSuggestionTitleUtils.ts
```diff
@@ -0,0 +1,24 @@
+import * as StringUtils from '../../utils/StringUtils';
+
+/*
+* Formats title adding highlighted query if it matches
+* */
```
niedzielski, line 5: Typo: extra star and missing space indent.

src/components/typeahead-suggestion-title/TypeaheadSuggestionTitleUtils.ts
```diff
@@ -0,0 +1,24 @@
...
+export function getTitleChunks( query: string, title: string ): string[] {
```
niedzielski, line 6: Optional: you may want to return a tuple of strings like `[string, string, string]` and pad with empty strings as needed like `[ title, '', '' ]`. This would allow consumers to always operate on the same number of elements.

niedzielski, line 6: Optional: consider renaming to splitStringAtMatch.

src/components/typeahead-suggestion-title/TypeaheadSuggestionTitleUtils.ts
```diff
@@ -0,0 +1,24 @@
...
+	if ( matchStartIndex < 0 ) {
+		return [ StringUtils.htmlEscape( title ) ];
+	}
+
+	const matchEndIndex = matchStartIndex + sanitizedQuery.length;
+	const highlightedTitle = title.substring( matchStartIndex, matchEndIndex );
+	const beforeHighlight = title.substring( 0, matchStartIndex );
+	const afterHighlight = title.substring( matchEndIndex, title.length );
+
+	return [ beforeHighlight, highlightedTitle, afterHighlight ];
```
niedzielski, line 23: Do we need the HTML escaping or does Vue.js do that for us?

src/components/typeahead-suggestion/TypeaheadSuggestion.ts
speedandfunction-anribolon: Sure, i'll remove it

src/components/typeahead-suggestion-title/TypeaheadSuggestionTitleUtils.ts
speedandfunction-anribolon: Good idea @niedzielski !

src/components/typeahead-suggestion-title/TypeaheadSuggestionTitleUtils.ts
speedandfunction-anribolon: Right, I'll remove

src/components/typeahead-suggestion/TypeaheadSuggestion.vue
```diff
@@ -0,0 +1,131 @@
...
+		box-shadow: 0 0 @border-width-base @border-width-base @wmui-color-base80;
+		border-radius: @border-radius-base;
+		background-position: center;
+		background-repeat: no-repeat;
+		// stylelint-disable-next-line plugin/no-unsupported-browser-features
+		background-size: cover;
+	}
+
+	&__thumbnail-placeholder {
+		// temporary placeholder, probably will be replaced with <wvui-icon>
```
niedzielski, line 112: Good note. Please add a "todo" as well so this will stand out more.

src/components/typeahead-suggestion/TypeaheadSuggestion.vue
```diff
@@ -0,0 +1,131 @@
...
+		}
+	}
+} );
+</script>
+
+<style lang="less">
+@import ( reference ) '@/themes/wikimedia-ui.less';
+
+.wvui-typeahead-suggestion {
+	background-color: #fff;
```
niedzielski, line 83: I'm unsure but maybe this should reference base100 instead of hardcoding white. /cc @Volker-E 

https://github.com/wikimedia/wikimedia-ui-base/blob/master/wikimedia-ui-base.less#L45

src/components/typeahead-suggestion-title/TypeaheadSuggestionTitle.vue
```diff
@@ -0,0 +1,54 @@
...
+		<!--eslint-disable-next-line max-len-->
+		{{ titleChunks[ 0 ] }}<em v-if="titleChunks[ 1 ]" class="wvui-typeahead-suggestion__match">{{ titleChunks[ 1 ] }}</em>{{ titleChunks[ 2 ] }}
+	</span>
+</template>
+
+<script lang="ts">
+import Vue, { PropType } from 'vue';
+import { splitStringAtMatch } from './TypeaheadSuggestionTitleUtils';
+
+export default Vue.extend( {
```
niedzielski, line 12: It's pretty neat how little JavaScript is necessary within the component itself. It's just stitching things together!

src/components/typeahead-suggestion/TypeaheadSuggestion.stories.ts
```diff
@@ -0,0 +1,112 @@
...
+					query="co"
+					:suggestion="suggestion"
+					:key="suggestion.id"
+				/>
+			</li>
+		</ol>
+		`
+	} );
+
+export const withInput = (): Vue.Component =>
```
niedzielski, line 69: I just tested this scenario and it looks great :+1: It looks the cursor keys aren't working though. Can you check it out?

.eslintignore
```diff
@@ -22,3 +22,6 @@
 
 # Ignore Jest snapshots.
 *.snap.ts
+
+# Ignore json files with mock data
```
Volker-E, line 26: JSON uppercase pls and `.` at eol

src/components/typeahead-suggestion-title/TypeaheadSuggestionTitle.vue
```diff
@@ -0,0 +1,54 @@
...
+} );
+</script>
+
+<style lang="less">
+@import ( reference ) '@/themes/wikimedia-ui.less';
+
+.wvui-typeahead-suggestion__title {
+	display: block;
+	margin: 0 0 2px 0;
+	color: @wmui-color-base10;
```
Volker-E, line 43: @wmui-colors shouldn't be used directly, only the @color-var itself. In this case it should be @color-base

src/components/typeahead-suggestion-title/TypeaheadSuggestionTitle.vue
```diff
@@ -0,0 +1,54 @@
...
+	font-size: @font-size-search-suggestion-title;
+	font-weight: bold;
```
Volker-E, line 45: We already feature a var for font-weight as well `@font-weight-bold`

src/components/typeahead-suggestion/TypeaheadSuggestion.vue
```diff
@@ -0,0 +1,131 @@
...
+	// stylelint-disable-next-line plugin/no-unsupported-browser-features
+	display: flex;
+	align-items: center;
+	padding: 8px @padding-horizontal-base;
+	text-decoration: none;
+
+	// css class is supposed to be used when navigating with keyboard
+	&:hover,
+	&--active {
+		background-color: @wmui-color-accent90;
```
Volker-E, line 93: Same as above, the `wmui-colors` shouldn't be referenced, only variables should reference them, here `background-color-typeahead-suggestion--hover`?

src/components/typeahead-suggestion/TypeaheadSuggestion.vue
```diff
@@ -0,0 +1,131 @@
...
+
+.wvui-typeahead-suggestion {
+	background-color: #fff;
+	// stylelint-disable-next-line plugin/no-unsupported-browser-features
+	display: flex;
+	align-items: center;
+	padding: 8px @padding-horizontal-base;
+	text-decoration: none;
+
+	// css class is supposed to be used when navigating with keyboard
```
Volker-E, line 90: It's better to name it than referencing it
"`&--active` is supposed to be used when navigating with keyboard."

src/components/typeahead-suggestion/TypeaheadSuggestion.vue
```diff
@@ -0,0 +1,131 @@
...
+	&:hover,
+	&--active {
+		background-color: @wmui-color-accent90;
+	}
+
+	&__thumbnail-placeholder,
+	&__thumbnail {
+		display: block;
+		min-width: @thumb-width-search-suggestion;
```
Volker-E, line 99: Why are we putting `min-width` and `max-width` with the same values here? Instead of just  width?

src/components/typeahead-suggestion/TypeaheadSuggestion.vue
```diff
@@ -0,0 +1,131 @@
...
+		min-height: @thumb-height-search-suggestion;
+		max-width: @thumb-width-search-suggestion;
+		max-height: @thumb-height-search-suggestion;
+		box-shadow: 0 0 @border-width-base @border-width-base @wmui-color-base80;
```
Volker-E, line 103: This should prob use border  instead of `box-shadow` or what's the limitation overcome?

src/themes/wikimedia-ui.less
```diff
@@ -21,3 +21,8 @@
 @box-shadow-widget: inset 0 0 0 1px transparent;
 
 @size-indicator: 12 / @font-size-browser / @font-size-base;
+
+@thumb-width-search-suggestion: 34px;
```
Volker-E, line 25: width-search-suggestion-thumb. That should prob be set in `em`s as well to be resizable on text-zoom

src/components/typeahead-suggestion/TypeaheadSuggestion.vue
speedandfunction-anribolon: @Volker-E fait remark, but this approach was [suggested](https://github.com/wikimedia/wvui/pull/68#discussion_r469289902) by @j4n-co and honestly saying I like it. But I'm opened to discussion for sure!

src/components/typeahead-suggestion/TypeaheadSuggestion.vue
speedandfunction-anribolon: fair enough :)

src/components/typeahead-suggestion/TypeaheadSuggestion.vue
speedandfunction-anribolon: Sure I'll change it

src/components/typeahead-suggestion/TypeaheadSuggestion.vue
Volker-E: Indeed, very fancy. Then please ignore my question. It'd be useful to add a comment though above box-shadow hinting at the reason as this is no too common pattern.

src/components/typeahead-suggestion/TypeaheadSuggestion.vue
niedzielski: Sounds like this can be wvui-icon now :+1:

Volker-E:
> As @niedzielski mentioned in Slack, there was one issue with the titles which was resolved by truncating the titles onto one line.
> image-2
> This issue is caused by my earlier suggestion of using text-indent instead of padding to achieve a RTL/LTR agnostic padding between the thumbnail and the text. I did not consider multiple lines in that scenario.
> 
> However, I don't think we should truncate the titles. I think that could be point of contention for editors, because titles are very important when searching.
> 
> I was thinking of what other options we have other than truncating the titles, I'm leaning towards # 3 :
> 
>     1. flexbox `gap` property - Barely any browser support.
> 
>     2. `padding-block-start` - Doesn’t work in IE11 but does on new modern browsers.
> 
>     3. Insert a blank `<span>` between the thumbnail and text, with a min-width of 12px. It's hacky, but I think it’s a reasonable stand-in for `gap` if we identify it as such e.g. `.flexbox-gap { min-width: 12px}`
> 
>     4. Switch back to normal padding and pretend this never happened.
> 
> 
> The reason I wouldn't ignore this issue altogether, and why it can't be solved by CSSJanus in MediaWiki, is that _if_ (and it is currently an _if_) we end up supporting language switching in search via a language-picker or something, then we will have to support client-side switching between RTL/LTR as well. If this scenario never materializes, then this could be a non-issue.
> 
> But I'd be interested If @Volker-E or anyone else has thoughts on this issue or objections to doing # 3.

Thanks for bringing this up @j4n-co, good rundown. Did you not take #2 into closer choice because of support only? We could hack around that with `@supports`.

nicholasray:
I'm actually surprised at `padding-block-start`'s  current level of browser support; that is something to look forward to 😀! 

However, according to caniuse Edge 12-18 also isn't supported [1] so it seems like we'd need to provide a fallback. If that's the case, I'm not sure if that option is better than #3 

[1] https://caniuse.com/#search=padding-block-start

j4n-co:
@Volker-E good suggestion with `@supports`, I hadn't thought of that. To @nicholasray's point, I guess it depends what the fallback is. I don't think we want the text & thumbnail to touch in any supported browser, so what if as a fallback, we add the extra padding to both the right and left side of the text container? That'll be RTL friendly (and hopefully not too visually awkward) and for newer browsers we'll add the extra padding only on the correct side and remove it from the other side?

nicholasray:
> @Volker-E good suggestion with `@supports`, I hadn't thought of that. To @nicholasray's point, I guess it depends what the fallback is. I don't think we want the text & thumbnail to touch in any supported browser, so what if as a fallback, we add the extra padding to both the right and left side of the text container? That'll be RTL friendly (and hopefully not too visually awkward) and for newer browsers we'll add the extra padding only on the correct side and remove it from the other side?

Good point, I was thinking the fallback would end up being a spacer span anyways, but I'd be down for the fallback to be symmetric padding. That sounds a lot less painful

speedandfunction-anribolon:
@Volker-E @j4n-co Do you  mean `padding-inline-start`?
https://developer.mozilla.org/en-US/docs/Web/CSS/padding-block-start
`padding-block-start: 12px` will add padding-top equal 12px (assuming that we have `writing-mode: horizontal-tb`)

j4n-co (approved):
Looks good to me! Thanks for all the updates. 

The CI error is just a bundlesize check.

niedzielski:
@Volker-E, I believe @speedandfunction-anribolon has addressed your feedback. Since you're out this week and there's a design offsite next week, I'm dismissing yoru GitHub review. Please follow up when you can with any concerns and we'll tackle them in a new PR! :+1:

PR #70, opened 2020-07-28
[component][typeahead-suggestion] Add highlighting as a directive

speedandfunction-anribolon:
I've decided to refactor highlighting functionality for a suggestion title (see [this patch](https://github.com/wikimedia/wvui/pull/68) ) and ended up with a directive.

### DIFF
--- a/src/components/typeahead-suggestion/TypeaheadSuggestion.test.ts
+++ b/src/components/typeahead-suggestion/TypeaheadSuggestion.test.ts
@@ -4,11 +4,15 @@ import { TypeaheadSuggestion } from './TypeaheadSuggestion';
 import * as suggestionsList from './TypeaheadSuggestion.stories.json';
 
 describe( 'matches the snapshot', () => {
-	type Case = [string, Record<string, TypeaheadSuggestion>];
+	type Case = [string, Record<string, TypeaheadSuggestion | string>];
 
 	const cases: Case[] = [
-		[ 'With thumbnail', { suggestion: suggestionsList.pages[ 1 ] as TypeaheadSuggestion } ],
-		[ 'Without thumbnail', { suggestion: suggestionsList.pages[ 0 ] as TypeaheadSuggestion } ]
+		[ 'With thumbnail', {
+			suggestion: suggestionsList.pages[ 1 ] as TypeaheadSuggestion, query: 'ob'
+		} ],
+		[ 'Without thumbnail', {
+			suggestion: suggestionsList.pages[ 0 ] as TypeaheadSuggestion, query: '132'
+		} ]
 	];
 
 	test.each( cases )( 'Case %# %s: (%p) => HTML', ( _, props ) => {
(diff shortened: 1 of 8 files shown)


src/directives/index.ts
```diff
@@ -0,0 +1,36 @@
...
+		const afterHighlight = title.substring( matchEndIndex, title.length );
+
+		// eslint-disable-next-line max-len
+		el.innerHTML = `${beforeHighlight}<em class="wvui-typeahead-suggestion__matching-title">${highlightedTitle}</em>${afterHighlight}`;
+	};
+
+/*
+* Highlights text if there are matches with a query
+* */
+export const highlightedText: DirectiveOptions = {
```
niedzielski, line 33: Would a namespace be better here?

src/directives/index.ts
```diff
@@ -0,0 +1,36 @@
+import WvuiUtils from '../utils/Utils';
```
niedzielski, line 1: Did you consider testing this directly instead of the trough a Vue.js component?

src/directives/index.ts
```diff
@@ -0,0 +1,36 @@
...
+import { DirectiveFunction, DirectiveOptions } from 'vue';
+
+const highlightedTextDirectiveHandler: DirectiveFunction =
+	( el: HTMLElement, binding ): void => {
```
niedzielski, line 5: If you define the type of the variable, `DirectiveFunction` in this case, I don't think you need to define the types of the parameters or return types as they should be correctly inferred.

src/directives/index.ts
```diff
@@ -0,0 +1,36 @@
...
+			return;
+		}
+
+		const matchEndIndex = matchStartIndex + sanitizedQuery.length;
+		const highlightedTitle = title.substring( matchStartIndex, matchEndIndex );
+		const beforeHighlight = title.substring( 0, matchStartIndex );
+		const afterHighlight = title.substring( matchEndIndex, title.length );
+
+		// eslint-disable-next-line max-len
+		el.innerHTML = `${beforeHighlight}<em class="wvui-typeahead-suggestion__matching-title">${highlightedTitle}</em>${afterHighlight}`;
```
niedzielski, line 27: Is this a good approach? We miss out on Singe File Components if we use an HTML string instead.

src/components/typeahead-suggestion/TypeaheadSuggestion.test.ts
```diff
@@ -4,11 +4,15 @@ import { TypeaheadSuggestion } from './TypeaheadSuggestion';
 import * as suggestionsList from './TypeaheadSuggestion.stories.json';
 
 describe( 'matches the snapshot', () => {
-	type Case = [string, Record<string, TypeaheadSuggestion>];
+	type Case = [string, Record<string, TypeaheadSuggestion | string>];
```
niedzielski, line 7: I think this would be more accurate to write out as a specific type `{suggestion: TypeaheadSuggestion, query: string}` instead of a generic Record.

src/components/typeahead-suggestion/TypeaheadSuggestion.vue
```diff
@@ -61,31 +67,6 @@ export default Vue.extend( {
 			return `/wiki/${this.suggestion?.key}`;
 		},
 		/*
-		* Formats title adding highlighted query if it matches
-		* */
-		suggestionTitle(): string | undefined {
```
niedzielski, line 66: Instead of breaking this out into a directive, did you consider splitting out a highlight component that is composed into the typeahead-suggestion component template?

PR #72, opened 2020-07-31
[component][typeahead-suggestion] Add network client for typeahead-search component

berndsi:
* Add fetch wrapper (only supports native fetch at this time)
* Add REST API search client

This network client provides fetchByTitle(), to be used for typeahead search suggestions.
Not providing accept-language header at this time since it does not seem to make a difference.

By default the requests are mocked, so no network is used.
If the environment variable TEST_LIVE_REQUESTS is set to a truthy value, the mocks are
disabled and requests are made against a live server.

Bug: T244287


src/http/fetch.ts
```diff
@@ -0,0 +1,24 @@
+// A wrapper for native fetch() in browsers.
+// Currently this rejects the returned promise if window.fetch is not available in the browser.
+// The plan is to add a fallback so we can support older browsers in the future.
+export function fetch( resource: string, init?: RequestInit ): Promise<Response> {
```
niedzielski, line 4: As you noted, we know we'll need to support IE11. How best to go about it? These are some of the concerns that come to mind:

- We want to be able to easily test either in implementation.
- This client is currently touching the Window global. State and especially globals can be problematic but is this ok?
- IE11 also doesn't support Promises.
- The library will be loaded and configured once but may be depended upon by multiple codebases, not just Vector.

Did you consider passing `window.fetch` / `$.ajax` / custom Fetch implementation as a dependency of the SearchClient, along with query and host, instead of creating this utility function? Passing fetch would allow you to avoid relying on the `window` global and allow to test any implementation you wanted.

src/http/fetch.test.ts
```diff
@@ -0,0 +1,95 @@
+import { buildQueryString, fetch } from './fetch';
+import * as jestFetchMock from 'jest-fetch-mock';
+
+jestFetchMock.enableFetchMocks();
+const mockedRequests = !process.env.TEST_LIVE_REQUESTS;
+
+describe( 'fetch() using window.fetch', () => {
+	const url = '//en.wikipedia.org/w/rest.php/v1/search/title?q=jfgkdajgioj&limit=10';
+
+	jestFetchMock.enableFetchMocks();
```
niedzielski, line 10: Should this be in the beforeEach()?

src/http/fetch.test.ts
```diff
@@ -0,0 +1,95 @@
...
+
+	beforeEach( () => {
+		fetchMock.resetMocks();
+		if ( !mockedRequests ) {
+			fetchMock.disableMocks();
+		}
+		fetchMock.mockIf( /^\/\/en.wikipedia.org\//, async ( req ) => {
```
niedzielski, line 17: Do we care about the language subdomain here?

src/http/fetch.test.ts
```diff
@@ -0,0 +1,95 @@
+import { buildQueryString, fetch } from './fetch';
+import * as jestFetchMock from 'jest-fetch-mock';
+
+jestFetchMock.enableFetchMocks();
+const mockedRequests = !process.env.TEST_LIVE_REQUESTS;
+
+describe( 'fetch() using window.fetch', () => {
```
niedzielski, line 7: Do you think we need these tests? I'm unsure. When it's not mocked, we test the actual Fetch implementation but that's going to be either window.fetch() or $.ajax(), neither of which is ours.

src/components/typeahead-search/http/restApiSearchClient.ts
```diff
@@ -0,0 +1,70 @@
...
+	};
+}
+
+// https://www.mediawiki.org/wiki/API:REST_API/Reference#Autocomplete_page_title
+function fetchByTitle(
+	searchText: string,
+	domain: string,
+	limit: number
+): Promise<SearchResponse> {
+	const params: Record<string, string|number> = {
```
niedzielski, line 50: Do these locals need explicit typing or can they be inferred?

src/components/typeahead-search/http/restApiSearchClient.ts
```diff
@@ -0,0 +1,70 @@
...
+interface RestResult {
+	id: number;
+	key: string;
+	title: string;
+	excerpt?: string;
+	description?: string;
+	thumbnail?: RestThumbnail;
+}
+
+interface RestThumbnail {
```
niedzielski, line 19: A number of these fields appear to be nullable unfortunately.

https://www.mediawiki.org/wiki/API:REST_API/Reference#Schema

src/components/typeahead-search/http/restApiSearchClient.test.ts
```diff
@@ -0,0 +1,101 @@
+import { restSearchClient } from './restApiSearchClient';
+import * as jestFetchMock from 'jest-fetch-mock';
+
+jestFetchMock.enableFetchMocks();
+const mockedRequests = !process.env.TEST_LIVE_REQUESTS;
+
+describe( 'restSearchClient', () => {
+	beforeEach( () => {
```
niedzielski, line 8: Do we need that enable mocks thingy here?

src/http/fetch.test.ts
berndsi: Not really. I just wanted to use a request that has a chance of working live as well and provides JSON response. I preferred this because it's something we have some control over and not using servers of another organization.

src/http/fetch.test.ts
berndsi: Originally I had this hitting example.org and had them as more general fetch calls. Now that I switched to something very similar to what the search client does the 200 response tests do seem redundant. I'd probably still prefer to keep the 404 test. I was actually considering adding some more tests for network issues since the jQuery version does behave quite differently. (It throws exceptions, uses different fields than fetch.)

src/components/typeahead-search/http/restApiSearchClient.ts
berndsi: Good catch. I'll remove it.

src/components/typeahead-search/http/restApiSearchClient.ts
berndsi: Good catch! Will push that update soon.

src/components/typeahead-search/http/restApiSearchClient.test.ts
berndsi: It was up there on line 4 but I'm moving it to a beforeAll() to be consistent with fetch.test.ts.

src/http/fetch.ts
berndsi: Support for IE11 is tough here without adding to the bundle size. Not sure how much bloat to the bundle size is acceptable here.
Passing in a fetch implementation seems problematic since `window.fetch` and `$.ajax` behave slightly differently. We would have to wrap the `$.ajax` call if we wanted to make it behave like fetch.
IE11 not supporting Promises throws in another wrench. Makes me wonder if I should drop fetch for now and go with callbacks and XHR. (So far I only see await used in a few tests, beside what this PR adds.)
I wonder if it's worth doing a prototype using a ponyfill for Promises and fetch at all, to see how much it adds to the bundle size.

niedzielski: I think the jQuery promises are actually ES6 compatible? I think it'd be ok if Vector unconditionally passed $.ajax in if that simplifies things. We just shouldn't assume the fetch implementation in the library.

https://jquery.com/upgrade-guide/3.0/#breaking-change-and-feature-jquery-deferred-is-now-promises-a-compatible
https://www.mediawiki.org/wiki/JQuery

niedzielski: If I understand correctly, fetch is not a requirement we can demand due to the support requirements. Our options are:

1. Ask the client to pass in a fetch implementation (likely to be jQuery). [Unofficial note on some differences](https://stackoverflow.com/a/47623125/970346).
2. Pony/polyfill fetch _and_ Promise. This will increase our bandwidth usage and may still behave differently from native fetch. For instance, see these [caveats for the Preact guy / MDN's implementation](https://github.com/developit/unfetch#caveats).
3. Use XHR. Callbacks but on the plus side we cancel requests.
4. Something else?

@speedandfunction-anribolon, thoughts on using an XMLHttpRequest?

src/http/fetch.test.ts
berndsi: I've moved it into beforeAll().

src/components/typeahead-search/http/SearchClient.ts
```diff
@@ -0,0 +1,37 @@
...
+	 * @param limit maximum number of results
+	 */
+	fetchByTitle(
+		searchText: string,
+		domain: string,
+		limit?: number
+	): Promise<SearchResponse>;
+}
+
+export interface SearchResponse {
```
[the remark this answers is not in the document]
berndsi: Are you talking about the searchText string? What would this be used for?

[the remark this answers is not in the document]
niedzielski: This is used to highlight any matching text in the result titles. You can see this on wikipedia.org if you type "ba". The Banana article title will be emphasized like **Ba**nana.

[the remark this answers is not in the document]
berndsi: Ok, if this makes it easier for the client then that's good, and I'll add it.

src/components/typeahead-search/http/restApiSearchClient.test.ts
```diff
@@ -0,0 +1,106 @@
...
+	} );
+
+	afterAll( () => {
+		jestFetchMock.disableFetchMocks();
+	} );
+
+	beforeEach( () => {
+		fetchMock.resetMocks();
+		if ( !mockedRequests ) {
+			fetchMock.disableMocks();
```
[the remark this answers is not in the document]
berndsi: The motivation for the one in `afterAll()` was to undo what was done in `beforeAll()`.

Not exactly sure what's going on here. The comment in the d.ts file says
> // alias of fetchMock.disableMocks() for ease of ES6 import syntax to not clash with other libraries

When I try using just `fetchMock.disableMocks()` instead of `jestFetchMock.disableFetchMocks()` that works but we lose the symmetry of beforeAll and afterAll. When also replacing the `enableMocks` call it forces me to change the jestFetch mock import. I tried just `import 'jest-fetch-mock'` but that fails the tests: `ReferenceError: fetchMock is not defined`.

src/components/typeahead-search/http/restApiSearchClient.test.ts
```diff
@@ -0,0 +1,106 @@
...
+		}
+	} );
+
+	test( '2 results', async () => {
+		const thumbUrl = '//upload.wikimedia.org/wikipedia/commons/0/01/MediaWiki-smaller-logo.png';
+		fetchMock.mockResponse( JSON.stringify( {
+			pages: [
```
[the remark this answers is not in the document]
berndsi: How about `const mockedResponse: RestResponse = {...` and `fetchMock.mockResponse( JSON.stringify( mockedResponse ) );` since RestResponse is already exported?

src/components/typeahead-search/http/restApiSearchClient.ts
```diff
@@ -0,0 +1,69 @@
...
+						width: page.thumbnail.width,
+						height: page.thumbnail.height
+					} : undefined
+				} ) )
+	};
+}
+
+// https://www.mediawiki.org/wiki/API:REST_API/Reference#Autocomplete_page_title
+function fetchByTitle(
+	searchText: string,
```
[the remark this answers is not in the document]
berndsi: That sounds like a great idea. Do you think we also need to add validation of the limit parameter, to be within the 1 and 100 range?

src/components/typeahead-search/http/restApiSearchClient.test.ts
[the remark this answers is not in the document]
niedzielski: Sure, I think that makes sense.

src/components/typeahead-search/http/restApiSearchClient.test.ts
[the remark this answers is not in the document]
niedzielski: I haven't used jest-fetch-mock (or at least don't remember having used it). I'm looking at the docks and wondering if you considered using `mockOnce()` or `mockOnceIf()`. It seems like those might be cleaner.

src/components/typeahead-search/http/restApiSearchClient.ts
```diff
@@ -0,0 +1,69 @@
...
+	domain: string,
+	limit: number
+): Promise<SearchResponse> {
+	const params = {
+		q: searchText,
+		limit: limit
+	};
+	const headers: Record<string, string> = {
```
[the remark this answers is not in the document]
berndsi: Removed it.

src/components/typeahead-search/http/restApiSearchClient.ts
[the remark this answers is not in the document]
berndsi: Added skipping the request if searchText is empty. Yeah, the server already checks that. So, not needed here. I've started looking into adding a test for error responses but ran into some issues.

src/components/typeahead-search/http/restApiSearchClient.test.ts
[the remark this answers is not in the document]
berndsi: Done

src/components/typeahead-search/http/restApiSearchClient.test.ts
```diff
@@ -0,0 +1,117 @@
+import { RestResponse, restSearchClient } from './restApiSearchClient';
+import * as jestFetchMock from 'jest-fetch-mock';
+
+const mockedRequests = !process.env.TEST_LIVE_REQUESTS;
+
+describe( 'restSearchClient', () => {
```
niedzielski, line 6: I think this should match the test subject.

src/components/typeahead-search/http/restApiSearchClient.ts
```diff
@@ -0,0 +1,72 @@
...
+		results:
+			restResponse.pages
+				.map( ( page ) => ( {
+					id: page.id,
+					key: page.key,
+					title: page.title,
+					description: page.description,
+					thumbnail: page.thumbnail ? {
+						url: page.thumbnail.url,
+						width: page.thumbnail.width,
```
niedzielski, line 36: You may want to set width and height to zero if falsy as they may be null according to the docs (and presumably undefined as well):

  width: page.thumbnail.width || 0,
  height: page.thumbnail.height || 0

src/components/typeahead-search/http/restApiSearchClient.ts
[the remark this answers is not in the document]
niedzielski: Sorry, I really should have said _blank_ not empty I think it would be reasonable to assume that searchText always has leading and trailing whitespace removed:

  searchText = searchText.trim()

If this is done before the length is tested, it will fix the searchText.length test below.

Additionally, since an empty string is falsy, I think writing `if ( !searchText ) { ... }` would do just as well.

src/http/fetch.ts
speedandfunction-anribolon: @niedzielski @berndsi I wouldn't use XMLHttpRequest in spite of it's a bulletproof solution. I'd better stick with polyfills. Maybe we can deliver separate bundles for IE including polyfills for fetch and promises and "polifill-less" bundle for the rest of browsers?

src/components/typeahead-search/http/restApiSearchClient.ts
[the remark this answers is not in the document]
berndsi: Good point. Fixed it.

src/components/typeahead-search/http/restApiSearchClient.test.ts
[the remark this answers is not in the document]
berndsi: The docs of `mockOnce()` seems to imply that it always mocks:

> but guarantees the next call to fetch will be mocked even if the default behavior of fetchMock is to use the real implementation

I think it would be nice to have the option to run these tests against a real server to keep these grounded in reality, and not just make up some responses that are unlikely to happen in the real world.

src/components/typeahead-search/http/restApiSearchClient.test.ts
berndsi: Done

src/components/typeahead-search/http/restApiSearchClient.ts
berndsi: Is `0` the correct value or should we just make it undefined?

niedzielski: How about `page.thumbnail.width ?? 0`?

berndsi: If we make the thumbnail width or height 0 then the thumbnail would not be shown. Let's consider the following HTML:

```
There should be an image:
<img width="0" src="https://upload.wikimedia.org/wikipedia/commons/thumb/2/2b/Jupiter_and_its_shrunken_Great_Red_Spot.jpg/200px-Jupiter_and_its_shrunken_Great_Red_Spot.jpg"/>
```

I wonder if it would be better to just not provide a width (or height) if the REST API has null values for width (or height respectively). The UI component probably would want to set a maxWidth and maxHeight.

niedzielski: Oh, sorry my mistake. How about `page.thumbnail.width ?? undefined`? This means use any number the server sends, otherwise use undefined.

berndsi: Sounds good to me. Still getting used to the nullish coalescing operator. I've updated the code.

src/components/typeahead-search/http/restApiSearchClient.test.ts
```diff
@@ -0,0 +1,117 @@
...
+		const searchResult = await restSearchClient().fetchByTitle(
+			'media',
+			'en.wikipedia.org',
+			2
+		);
+
+		expect( searchResult.searchText ).toEqual( 'media' );
+		expect( searchResult.results ).toBeTruthy();
+		expect( searchResult.results.length ).toBe( 2 );
+		expect( searchResult.results[ 0 ] ).toStrictEqual( {
```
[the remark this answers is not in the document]
berndsi: Good idea. It doesn't work 1:1 (the null thumbnail changes to undefined) but I've clarified that in the latest version of the test.

src/components/typeahead-search/http/restApiSearchClient.test.ts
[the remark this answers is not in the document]
niedzielski: I see but wouldn't this move to just before the call to `fetchByTitle()` in the tests themselves? For example, `fetch.doMockOnceIf( mockedRequests )`.

This isn't a big deal on it's own but I wanted to encourage you to take the time to write the best network tests here as possible since others contributors are likely to copy your approach. At the other extreme, I remember in the Android app we had trouble with tests depending on other tests running in a certain order or being unable to run in parallel because of configuration changes to globals. I think the before/after parity you've given here will help us avoid those issues.

Then again, maybe all this breaks if change the underlying implementation from fetch to XHR :sob:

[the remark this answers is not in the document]
berndsi: @niedzielski I'm not sure I understand the first paragraph of your most recent comment. `fetchMock.mockResponse()` is right before the call to `fetchByTitle()`. Are you asking to just replace it with `fetch.doMockOnceIf()` or something else?

Yes, if we change to implementation away from fetch this all seems mood.

src/http/fetch.ts
berndsi: @niedzielski Any concerns using a polyfill bundle for IE11?

niedzielski: I think the polyfill will need to provide Promises as well. Do you have one in mind? How big is it?

niedzielski: I haven't used this but [es6-promise](https://github.com/stefanpenner/es6-promise#downloads) is [about 2.4 kB](https://bundlephobia.com/result?p=es6-promise@4.2.8) and [unfetch](https://github.com/developit/unfetch) is [about .5 kB](https://bundlephobia.com/result?p=unfetch@4.1.0). Thoughts? /cc @speedandfunction-anribolon

berndsi: I was also considering the `'core-js/modules/es6.promise'` part of core-js, as described in [Loading Polyfills Only When Needed](http://philipwalton.com/articles/loading-polyfills-only-when-needed/). The alternative for unfetch was [whatwg-fetch](https://www.npmjs.com/package/whatwg-fetch) but that is quite a bit larger at [3.1KB](https://bundlephobia.com/result?p=whatwg-fetch@3.3.0).
Before we drill down more into various options I thought there was a more general concern about having an extra bundle for older browser, was there not?

niedzielski: Let's move this discussion to Slack so Web can weigh in.

berndsi: @niedzielski Should we merge the fetch implementation, then create a separate PR for XMLHttpRequest? That way we could revert the latter commit to get back to a pure fetch implementation. If this is the case, are there any other open issues with this PR?

Side note: there is now a new Phab task to add a [fetch polyfill to mediawiki core](https://phabricator.wikimedia.org/T260198).

src/components/typeahead-search/http/restApiSearchClient.test.ts
```diff
@@ -0,0 +1,120 @@
...
+		};
+		fetchMock.mockOnce( JSON.stringify( restResponse ) );
+
+		const searchResult = await restSearchClient().fetchByTitle(
+			'media',
+			'en.wikipedia.org',
+			2
+		);
+
+		expect( searchResult.searchText ).toEqual( 'media' );
```
niedzielski, line 54: Any reason not to use toStrictEqual? I saw a couple places where equal was used and I wasn't sure if it was intentional.

src/components/typeahead-search/http/SearchClient.ts
```diff
@@ -0,0 +1,38 @@
+export interface SearchClient {
+	/**
+	 * @param searchText search string to search for.
+	 * @param domain the base URL for the wiki without protocol. Example: 'sr.wikipedia.org'
```
niedzielski, line 4: Nice example input.

src/components/typeahead-search/http/restApiSearchClient.test.ts
```diff
@@ -0,0 +1,120 @@
...
+	test( '0 results', async () => {
+		const restResponse: RestResponse = { pages: [] };
+		fetchMock.mockOnce( JSON.stringify( restResponse ) );
+
+		const searchResult = await restSearchClient().fetchByTitle(
+			'thereIsNothingLikeThis',
+			'en.wikipedia.org'
+		);
+
+		expect( searchResult.searchText ).toEqual( 'thereIsNothingLikeThis' );
```
niedzielski, line 82: Extreme nitpick and optional: consider using "query" terminology everywhere instead of "searchText". I don't have a good reason other than it seems a little more natural (to me) and succinct/

src/components/typeahead-search/http/restApiSearchClient.test.ts
berndsi: No good reason. Changed it.

src/components/typeahead-search/http/restApiSearchClient.test.ts
berndsi: Sounds good to me.

src/components/typeahead-search/http/restApiSearchClient.test.ts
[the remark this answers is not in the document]
berndsi: I've changed this to use mockOnce now.

src/components/typeahead-search/http/restApiSearchClient.ts
```diff
@@ -0,0 +1,73 @@
+import { buildQueryString, fetch } from '../../../http/fetch';
```
berndsi, line 1: @niedzielski Do you know why changing this to `'@/http/fetch'` doesn't work? I'm getting

```
    Cannot find module '@/http/fetch' from 'src/components/typeahead-search/http/restApiSearchClient.ts'

    Require stack:
      src/components/typeahead-search/http/restApiSearchClient.ts
      src/components/typeahead-search/http/restApiSearchClient.test.ts

    > 1 | import { buildQueryString, fetch } from '@/http/fetch';
        | ^
      2 | import { SearchClient, SearchResponse } from './SearchClient';
      3 | 
      4 | // https://www.mediawiki.org/wiki/API:REST_API/Reference#Search_result_object

      at Resolver.resolveModule (node_modules/jest-resolve/build/index.js:308:11)
      at Object.<anonymous> (src/components/typeahead-search/http/restApiSearchClient.ts:1:1)
```

src/http/fetch.ts
niedzielski: Merging with the expectation that we will use XHR or polyfill ASAP.

PR #73, opened 2020-08-01
[WIP][demo] add "Hello World"

niedzielski:
@j4n-co, not done yet but we can add one for experimenting with SSR client hydration since the Vector integration patch is currently blocked.


demos/hello-world/hello-world.html
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
+<html lang="en" dir="ltr">
+	<head>
+		<meta charset="utf-8" />
```
Volker-E, line 4: No self-closing tag `/` in HTML5!

Volker-E:
SSR is on future plan for Vue.js Migration team. But tackled as early as a few months from now and probably with a wider-reaching solution.

PR #76, opened 2020-08-18
[component][input] Add rtl support for input

speedandfunction-anribolon:
Add RTL support for input component.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -4,6 +4,7 @@ Versions and bullets are arranged chronologically from latest to oldest.
 
 ## v0.0.2 (unreleased)
 
+-   [component][input] Add rtl support for input
 -   [component][input] Add button control for input
 -   [component][typeahead-search] Add network client
 -   [docs][dev] Document way to clear Jest cache
--- a/src/components/input/Input.stories.ts
+++ b/src/components/input/Input.stories.ts
@@ -6,6 +6,19 @@ import WvuiButton from '../button/Button.vue';
 import { InputType } from './InputType';
 import './Input.stories.less';
 
+const directions = [ 'rtl', 'ltr' ];
+
+const directionMixin = {
+	watch: {
+		direction: ( newValue: string ) => {
+			window.document.documentElement.setAttribute( 'dir', newValue );
+		}
+	},
+	beforeMount() {
+		window.document.documentElement.dir = 'ltr';
+	}
+};
+
 export default {
 	title: 'Components/Input',
 	component: WvuiInput,
@@ -15,13 +28,18 @@ export default {
 export const configurable = (): Vue.Component =>
 	Vue.extend( {
 		components: { WvuiInput },
+		mixins: [ directionMixin ],
 		props: {
 			disabled: { type: Boolean, default: boolean( 'Disabled', false ) },
 			type: {
 				type: String as PropType<keyof typeof InputType>,
 				default: select( 'Input Type', Object.keys( InputType ), 'Search' )
 			},
-			placeholder: { type: String, default: text( 'Placeholder', 'Search…' ) }
+			placeholder: { type: String, default: text( 'Placeholder', 'Search…' ) },
+			direction: {
+				type: String,
+				default: select( 'Lang Direction', directions, 'ltr' )
+			}
 		},
 		data() {
 			return {
@@ -52,8 +70,13 @@ export const configurable = (): Vue.Component =>
 export const withStartIcon = (): Vue.Component =>
 	Vue.extend( {
 		components: { WvuiInput },
+		mixins: [ directionMixin ],
 		props: {
-			disabled: { type: Boolean, default: boolean( 'Disabled', false ) }
+			disabled: { type: Boolean, default: boolean( 'Disabled', false ) },
+			direction: {
+				type: String,
+				default: select( 'Lang Direction', directions, 'ltr' )
+			}
 		},
 		data() {
 			return {
@@ -76,8 +99,13 @@ export const withStartIcon = (): Vue.Component =>
 export const withEndIcon = (): Vue.Component =>
 	Vue.extend( {
 		components: { WvuiInput },
+		mixins: [ directionMixin ],
 		props: {
-			disabled: { type: Boolean, default: boolean( 'Disabled', false ) }
+			disabled: { type: Boolean, default: boolean( 'Disabled', false ) },
+			direction: {
+				type: String,
+				default: select( 'Lang Direction', directions, 'ltr' )
+			}
 		},
 		data() {
 			return {
@@ -99,8 +127,13 @@ export const withEndIcon = (): Vue.Component =>
 export const withClearAction = (): Vue.Component =>
 	Vue.extend( {
 		components: { WvuiInput },
+		mixins: [ directionMixin ],
 		props: {
-			disabled: { type: Boolean, default: boolean( 'Disabled', false ) }
+			disabled: { type: Boolean, default: boolean( 'Disabled', false ) },
+			direction: {
+				type: String,
+				default: select( 'Lang Direction', directions, 'ltr' )
+			}
 		},
 		data() {
 			return {
@@ -127,8 +160,13 @@ export const withClearAction = (): Vue.Component =>
 export const withButton = (): Vue.Component =>
 	Vue.extend( {
 		components: { WvuiInput, WvuiButton },
+		mixins: [ directionMixin ],
 		props: {
-			disabled: { type: Boolean, default: boolean( 'Disabled', false ) }
+			disabled: { type: Boolean, default: boolean( 'Disabled', false ) },
+			direction: {
+				type: String,
+				default: select( 'Lang Direction', directions, 'ltr' )
+			}
 		},
 		template: `
 		<div class="sb-input sb-input--has-button">
@@ -148,8 +186,11 @@ const searchLanguageMap = {
 	Japanese: '探す',
 	Greek: 'Αναζήτηση',
 	Swedish: 'Söka',
-	Mazandeerani: 'جستجو کردن'
+	Hebrew: 'לחפש',
+	Arabic: 'بحث'
 };
+
+const rtlWords = [ 'לחפש', 'بحث' ];
 
 export const wikipediaSearchInput = (): Vue.Component =>
 	Vue.extend( {
@@ -161,11 +202,22 @@ export const wikipediaSearchInput = (): Vue.Component =>
 				default: select( 'Label language', searchLanguageMap, 'Search' )
 			}
 		},
+		computed: {
+			placeholder() {
+				return `${this.buttonLabel}…`;
+			}
+		},
+		watch: {
+			buttonLabel( newValue ) {
+				const isRtl = rtlWords.indexOf( newValue ) >= 0;
+
+				document.documentElement.dir = isRtl ? 'rtl' : 'ltr';
+			}
+		},
 		template: `
 		<div class="sb-input sb-input--has-button">
 			<wvui-input
-				placeholder="Search…"
-				icon="search"
+				:placeholder="placeholder"
 				:disabled="disabled"
 				start-icon="test"
 				:clearable="true"
--- a/src/components/input/Input.vue
+++ b/src/components/input/Input.vue
@@ -107,7 +107,7 @@ export default Vue.extend( {
 			default: false
 		}
 	},
-	data() {
+	data(): Record<string, unknown> {
 		return {
 			currentValue: this.value,
 			// temporary hardcoded icons
@@ -287,5 +287,54 @@ export default Vue.extend( {
 	// For inline, inline-block, and table layouts.
 	vertical-align: middle;
 	user-select: none;
+}
+
+[ dir='rtl' ] {
+	.wvui-input {
+		& > .wvui-input__input {
+			direction: rtl;
+		}
+
+		&--has-start-icon {
+			.wvui-input__input {
+				padding-left: @padding-horizontal-input-text;
+				padding-right: @padding-horizontal-input-text * 2 + @size-icon;
+			}
+		}
+
+		&--has-end-icon {
+			.wvui-input__input {
+				padding-left: @padding-horizontal-input-text * 2 + @size-icon;
+				padding-right: @padding-horizontal-input-text;
+			}
+		}
+
+		// stylelint-disable-next-line no-descending-specificity
+		&__start-icon {
+			right: 0;
+			left: auto;
+			padding-right: @padding-horizontal-input-text;
+		}
+		// stylelint-disable-next-line no-descending-specificity
+		&__end-icon {
+			left: 0;
+			right: auto;
+			padding-right: @padding-horizontal-input-text;
+		}
+
+		&--clearable {
+			// add padding for input if it's clearable
+			.wvui-input__input {
+				padding-left: @size-icon + @padding-horizontal-input-text;
+			}
+		}
+
+		// Add right padding for clearable input
+		&--clearable.wvui-input--has-start-icon {
+			.wvui-input__input {
+				padding-right: @padding-horizontal-input-text * 2 + @size-icon;
+			}
+		}
+	}
 }
 </style>
--- a/bundlesize.config.json
+++ b/bundlesize.config.json
@@ -1,6 +1,6 @@
 {
 	"files": [
 		{ "path": "./dist/wvui.js", "maxSize": "2.3 KB" },
-		{ "path": "./dist/wvui.css", "maxSize": "1.1 KB" }
+		{ "path": "./dist/wvui.css", "maxSize": "1.2 KB" }
 	]
 }
--- a/src/components/input/Input.stories.less
+++ b/src/components/input/Input.stories.less
@@ -1,3 +1,5 @@
+@import ( reference ) '@/themes/wikimedia-ui.less';
+
 .sb-input {
 	width: 400px;
 
@@ -19,5 +21,17 @@
 			border-top-left-radius: 0;
 			border-bottom-left-radius: 0;
 		}
+	}
+}
+
+[ dir='rtl' ] .sb-input--has-button {
+	.wvui-input__input {
+		border-radius: 0 @border-radius-base @border-radius-base 0;
+		border-right-width: @border-width-base;
+		border-left-width: 0;
+	}
+
+	.wvui-button {
+		border-radius: @border-radius-base 0 0 @border-radius-base;
 	}
 }


src/components/input/Input.vue
```diff
@@ -288,4 +288,53 @@ export default Vue.extend( {
 	vertical-align: middle;
 	user-select: none;
 }
+
+[ dir='rtl' ] {
```
niedzielski, line 292: Did you consider using the bidir Less mixin approach we had talked about?

santhoshtr: Directionality support need to be introduced as part of this library so that the developers use helper css classes and you get direction agnostic styles generated automatically.  The CX UI library has them([Example](https://people.wikimedia.org/~santhosh/storybook/?path=/docs/layout-grid--padding-and-margin)). With such a framwork, we will never need to write padding, margin css. 

Not related to the framework, but in design, generally, I am a fan of symmetric spacing. i.e, I will try to avoid cases where padding-left != padding-right. But sometimes, like in the case of an icon and label, this may be unavoidable. We should stick to logical properties with the help of css classes there. No left and right, only start and end.

changelog.md
```diff
@@ -4,6 +4,8 @@ Versions and bullets are arranged chronologically from latest to oldest.
 
 ## v0.0.2 (unreleased)
 
+-   [component][input] Add rtl support for input
```
Volker-E, line 7: RTL uppercase pls

src/components/input/Input.snap.ts
```diff
@@ -32,87 +32,87 @@ exports[`matches the snapshot Case 1 Input Type: search : ({"type": "search"}) =
...
-        height="12"
+        height="20"
         role="presentation"
         viewBox="0 0 20 20"
-        width="12"
+        width="20"
         xmlns="http://www.w3.org/2000/svg"
       >
         <g
           fill="#72777d"
```
Volker-E, line 60: Haven't we got rid of non-black fill colors elsewhere already?

src/components/input/Input.snap.ts
```diff
@@ -32,87 +32,87 @@ exports[`matches the snapshot Case 1 Input Type: search : ({"type": "search"}) =
...
+    class="wvui-input__start-icon"
   >
     <span
       class="wvui-icon"
     >
       <svg
         aria-hidden="true"
-        height="12"
+        height="20"
         role="presentation"
```
Volker-E, line 54: no `role` when there's an `aria-hidden`

src/components/input/Input.stories.ts
```diff
@@ -14,13 +28,18 @@ export default {
...
 		components: { WvuiInput },
+		mixins: [ directionMixin ],
 		props: {
 			disabled: { type: Boolean, default: boolean( 'Disabled', false ) },
 			type: {
 				type: String as PropType<keyof typeof InputType>,
 				default: select( 'Input Type', Object.keys( InputType ), 'Search' )
 			},
-			placeholder: { type: String, default: text( 'Placeholder', 'Search...' ) }
+			placeholder: { type: String, default: text( 'Placeholder', 'Search…' ) },
```
Volker-E, line 38: multiline pls

src/components/input/Input.vue
```diff
@@ -276,4 +288,53 @@ export default Vue.extend( {
...
+		}
+		// stylelint-disable-next-line no-descending-specificity
+		&__end-icon {
+			left: 0;
+			right: auto;
+			padding-right: @padding-horizontal-input-text;
+		}
+
+		&--clearable {
+			// add padding for input if it's clearable
```
Volker-E, line 326: Self-explaining?

src/components/input/Input.vue
```diff
@@ -276,4 +288,53 @@ export default Vue.extend( {
...
+			.wvui-input__input {
+				padding-left: @size-icon + @padding-horizontal-input-text;
+			}
+		}
+
+		// Add right padding for clearable input
```
Volker-E, line 332: Self-explaining?

src/components/input/Input.vue
speedandfunction-anribolon: @santhoshtr I'm fond of the approach you're suggesting, but for now it's out of scope for us. @Volker-E I'd like you to participate in this discussion. For now I've implemented `bidi` LESS mixin as @niedzielski suggested. The question is - can we go further with `bidi` or should we reconsider RTL approach right now.

changelog.md
speedandfunction-anribolon: fixed

src/components/input/Input.snap.ts
speedandfunction-anribolon: it's outdated

src/components/input/Input.stories.ts
speedandfunction-anribolon: @Volker-E I'm not sure what you mean, could you please elaborate?

src/components/input/Input.test.ts
```diff
@@ -17,32 +17,34 @@ describe( 'matches the snapshot', () => {
 	} );
 } );
 
-it( 'should render an icon', () => {
-	const wrapper = shallowMount( WvuiInput, { propsData: { icon: 'search' } } );
+it( 'should render  start icon', () => {
```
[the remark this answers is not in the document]
speedandfunction-anribolon: fixed

edg2s:
CSSJanus support or similar is an absolute must requirement. We can't have a situation where we are expecting people to add and maintain RTL styles by hand, that would be a huge regression over our current development process.

src/components/input/Input.vue
edg2s: A padding & margin framework will only get you so far. More complex layouts will require absolute/relative positioning.

niedzielski:
@edg2s, any objections to deferring the CSSJanus discussion? It sounds like you and @santhoshtr have possibly different perspectives that will need to be reconciled and expect @j4n-co, @Volker-E, and others will want to weigh in as well. This will take time to get consensus and get right.

edg2s:
A bit. Temporary decisions have a way of become permanent, or growing exponentially as technical debt. Requiring manual RTL support instantly becomes a burden for any developer trying to add a component to this library. CSS flipping is a solved problem so there shouldn't be much to discuss. CSSJanus is written in JS with a PHP port for MediaWiki, and should be pretty easy to integrate into a build process. There is even an RTL plugin for postcss (postcss-rtl), although we should probably stick with Janus so we get bug-for-bug compatibility with MediaWiki.

Santhosh is pointing to a framework that should allow you to avoid using padding/margin in your stylesheets. Regardless of if you use that or not, you will still need RTL flipping for other CSS properties.

santhoshtr:
> Santhosh is pointing to a framework that should allow you to avoid using padding/margin in your stylesheets. Regardless of if you use that or not, you will still need RTL flipping for other CSS properties.

It is not just about padding/margin. It does not make sense using such margin/spacing helper classes without a grid framework, design vocabulary and a component library. If a developer is using such a framework and adding css with random borders, box shadow, text alignment, border radius etc, it defeat the purpose. My wish is WVUI being a library with a design language - including grid(flexbox and CSS grids already use logical directions), typography, spacing helpers, opinionated about borders, box-shadows etc, being expressed using a design vocabulary. In such a system, each building piece will have script directionality abstracted. In that case you won't be repeating a directionality specific style, hence you wont require CSS-Janus. The associated style will have both LTR and RTL(other directions too) styles rooted with `[dir=rtl`] kind of contexts.

CSS Janus has a postcss integration with [postcss-janus](https://github.com/elchininet/postcss-janus) library. As you can see the style output from it is not different from the framework I mentioned above. In a single CSS file, we get styles for both LTR and RTL.  

My preference for a smart framewrok and avoiding css-janus is because, framework can address bigger goals(mentioned above), it can progressively enhance to[ logical direction properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties) which is now supported by [larger set of browsers except IE11](https://caniuse.com/#feat=css-logical-props). This means, we should make way for that progressive enhancement in the framework and rely on logical properties. A system that use CSS-Janus does not cater that progressive enhancement. If we all rely on CSS-Janus flipping, we keep the culture of writing old CSS.

I think both me and Ed agree on the need to better directionality handling. The above discussion is only about how to do that in WVUI framework. It is a wider discussion depending on wishlist of WVUI beyong the vuejs-search project. I can totally understand the need to unblock vuejs-search with a minimal scope WVUI framework. The `.bidi()` approach used in this PR is in good direction, but I think the sooner we get layouting system in framework, the better.

edg2s:
> If a developer is using such a framework and adding css with random borders, box shadow, text alignment, border radius etc, it defeat the purpose.

We can still provide these helpers progressively and easily lint or review against people adding non-standard design language. For example once the padding and margin helpers are available, we can lint add padding and margin properties to our stylelint block-list. Realistically it is not possible to build a complex UI library with our desired browser support without occasionally using hand-coded position properties. See for example the SelectFileWidget in OOUI, where we have to use a near-transparent overlay widget with absolute position. That is unlikely to ever be solved by a framework.

Providing a framework to enforce design language is a great goal but is going to require some effort. CSSJanus on the other hand is ready now.

niedzielski:
Thanks all for the very nice discussion points above.

I don't think it's a question over whether CSSJanus is available or not. I think it's more that folks wish to deliberately pick the approach rather than inherit a default. The reason is the same as Ed already mentioned, temporary decisions stick. I wouldn't want to swim upstream against the overwhelming currents of the status quo.

The approach suggested in this patch may be suboptimal but it is contained and its shortcomings are clear. I'd like to move forward with it rather the alternatives of either 1) picking a permanent suboptimal solution in haste 2) blocking to have the lengthy discussion needed to pick the optimal solution. We already have the ticket [Ed made here](https://phabricator.wikimedia.org/T261323) which I think would be a great venue to continue the discussion in parallel. If the discussions resolves before shipping search, we'd be happy to use whatever approach participants come up with. If not, the shortcomings in the current approach will make the priority to choose a better alternative clear.

j4n-co:
Quick question, In practice, isn't CSSJanus going to be applied to these styles anyway once we integrate the components into mediaWiki and run them through ResourceLoader?

santhoshtr:
> Quick question, In practice, isn't CSSJanus going to be applied to these styles anyway once we integrate the components into mediaWiki and run them through ResourceLoader?

It is configurable in RL module definition. Use `"noflip": true` if the module already handle directionality.

edg2s:
> Quick question, In practice, isn't CSSJanus going to be applied to these styles anyway once we integrate the components into mediaWiki and run them through ResourceLoader?

I think that's possible. CC @catrope who knows more about RL + Vue integration.

If that is the case then RTL is already solved for all MediaWiki users, and unless we plan on prioritising non-MediaWiki users this early on (which seems like we'd be stretching ourselves thin), we can just ignore the problem for now.

catrope:
Yes, styles loaded through ResourceLoader are automatically run through CSSJanus unless they take pains to opt out.

niedzielski:
I don't think we're stretching to support third-parties at the moment but we do want development to make sense outside of the MediaWiki stack which this patch enables. Since using CSSJanus through the MediaWiki stack has the same tradeoffs as using it outside of the stack, that's not a good first choice since we'd inherit CSSJanus expectations. Additionally, the language selector is one of our next projects and changing the languages on the client is not possible using MediaWiki's CSSJanus without a full style reload. Figuring out those details is out of scope.

I would like to move forward with this patch to unblock search and, assuming we don't see any issues, opt-out of CSSJanus until it's decided as the path forward in [Ed's task](https://phabricator.wikimedia.org/T261323).

niedzielski:
> Surely the search component is going to be in MediaWiki where RTL already works, so why do you need to rush to add in hacky RTL support here?
I wouldn't say we're "rushing to add in hacky RTL support" but this is our burden to ship.

> Where do we expect to be able to change language without reloading the page?
This is being discussed for key UI in the [language selector project](https://www.mediawiki.org/wiki/Reading/Web/Desktop_Improvements). Additionally, I'll add that ContentTranslation ships both styles for their app. On disk, the size is terrible but it compresses extremely well.

edg2s:
> I wouldn't say we're "rushing to add in hacky RTL support" but this is our burden to ship.

Hand coding the RTL flips when we already have flipping technology is definitely a hack. It is a burden on every developer who comes along to add a component, not just the current team working on this library.

> This is being discussed for key UI in the [language selector project](https://www.mediawiki.org/wiki/Reading/Web/Desktop_Improvements).

That doesn't sound like a good thing to optimize for. Language switching is extremely rare as a percentage of page views. Trying to make it happen without reloading the page will introduce a huge amount of complexity (e.g. swapping out messages, language specific CSS which is not limited to directionality, and many other things) just to make an edge case slightly faster (and normal cases slower).

CX may be doing this in their experiment, but I would expect them not to continue doing this once it is integrated with ResourceLoader and properly deployed. Across the site loading extra RTL CSS for everything will add up on our bandwidth costs. CC @krinkle.

Krinkle:
> [logical direction properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties) which is now supported by [larger set of browsers except IE11](https://caniuse.com/#feat=css-logical-props).

Note that this "large set except IE 11" is "browsers released after March 2019 (iOS) or Jan 2020 (Edge+Opera)" thus doesn't even cover Grade A, and much less Grade C. We'll soon be raising Grade A's feature-test to require ES6 so this might be fine for new experimental code that is inherently post-interaction only, such as modal dialogs and ajax-y things. But, I expect the base design language and styles to not be that limited. The sooner we start coding that way, the less we'll have to remember/rewrite/refactor later on, and the less we'll be battling over time allocation to do all that work which might be hard as someone looking in might think we already delivered it, when in fact we didn't.

Software is an exciting area to work in, and part of that is solving problems within the reality we live in. Browser vendors are working hard to erase mistakes made in the past by including more into the platform every year which sometimes unlocks new things, sometimes makes solved problems even easier to solve, and sometimes adds confusing things that will never see adoption.

For things we've already solved and can't use two approaches side-by-side for, we need to wait until native capabilities are ubiquitous enough that we can adopt them within the scope of the Foundation's mission and not needlessly cut off audiences. At that time we'll surely consider allocating some additional resources (temporarily) to switch over to the new thing; prioritised based on how much benefit is to be had from that. For CSS logical-props, that time appears to not yet be upon us.

In any event, this problem in particular has already been solved in the industry over a decade ago. If we want native things to be nice from the get-go, we should pay close attention to upcoming specs and help iterate during the spec incubation, and origin trials; both for better end-user experience, localisation, performance, and dev ergonomics. The perfomance team pushed for Wikimedia to become a W3C member, which we now are, so this kind of participation while open to all, is even easier for WMF staff especially.

I think automony and experimentation are important. Past work on tooling considerations and switches  demonstrate that we're generally better off when that is allowed and encouraged. But, it's equally important that such experiments have boundaries, are handled responsibly, and well-communicated. To me that means they are best done in either the critical path of a project where that experiment is core to the (or few) purposes of the project, or done far outside the critical path where established part(s) are exchanged to explore unknowns.

It seems to me that directionality isn't a core problem of WVUI, so even if a new approach were ready for adoption, I'd advise against its use here. There is enough greenfield exploration here already that further widening might  prevent the project from coming to fruition.

Once we have a few good iterations released and behind us, this could be an interesting experiment to take on, in the following year?

Krinkle:
> CX may be doing this in their experiment, but I would expect them not to continue doing this once it is integrated with ResourceLoader and properly deployed.

Indeed. I only briefly looked at the CX branch, but it seems to not follow what was agreed in the FAWG; both in general approach and in scope (pilot scope is desktop refresh, not CX). I guess it was done this way to get a headstart on porting to Vue while we wait for other parts to be ready, but the specific way it currently works is not deployable. (Note that the RL infra that Roan worked on is now ready for use, afaik.)

> Across the site loading extra RTL CSS for everything will add up on our bandwidth costs. CC @Krinkle.

Aye, that'd be nice to avoid. In general I expect that to be small and compress well, though, so it's not a deal breaker for me. It will likely make only a negligable difference in size. In terms bandwidth-bytes vs time-spent, CSS tends to be cheap to parse for unused rules. JS on the other hand tends to be more expensive.

But yeah, as other comments indicated already, CSSJanus will naturally apply to this, so I don't think there's anything we need to do here. Just write CSS normally for LTR, and CSSJanus will do the rest.

If you do want to have an RTL demo as part of this repo as well, then one could use our cssjanus package or postcss-janus to produce an RTL version.

mooeypoo:
> I don't think we're stretching to support third-parties at the moment but we do want development to make sense outside of the MediaWiki stack which this patch enables. Since using CSSJanus through the MediaWiki stack has the same tradeoffs as using it outside of the stack, that's not a good first choice since we'd inherit CSSJanus expectations. Additionally, the language selector is one of our next projects and changing the languages on the client is not possible using MediaWiki's CSSJanus without a full style reload. Figuring out those details is out of scope.

I have to admit, I am also not sure why you'd want to purposefully put yourselves in the position of manually fighting RTL correction online. CSSJanus, as "old style" as it may be, is a result of over 10 years of learning from our mistakes and figuring out how to do stuff. There are other libraries out there today that may be slightly better (potentially) but doing it from scratch is something I'd **highly** recommend against. I don't know if I can emphasize just how terrible RTL behavior online already is, and how much you'd really have to climb the high spike-filled walls of internet-tries-to-RTL. I am not sure you realize just how much energy this would take -- and at the expense of solutions that are already solved.

I'd seriously avoid getting into that unless you **really really** have to.

And speaking of having to, I'm actually a little concerned about the idea of having a language switcher _without reloading the page_. This isn't a good idea, for multiple reasons, but the main one is that CSS isn't the only thing that controls RTL positioning (and the multiple hacks we need to do to enable RTL users) on an internet page.

There are a bunch of behaviors that rely on knowing the directionality of the page (or the segment they're in) on load. Some of those are:

- Some browser behavior is impacted by the directionality of the entire page, and switching it dynamically doesn't always refresh the browser behavior. 
- RTL behavior is not just CSS. There are **a lot** of Javascript corrections and behavior that we have to implement that depends on the direction of the page. You'd need to either make all of that reactive (eh...) or reload the components anyways.
- There are also cases where the order of HTML elements *matters* in RTL. Those may be a bit edge casey, but half of RTL support is Edge Case unfortunately as it is. This means that dynamically switching the language may break on those.

Users don't usually switch languages very often back-and-forth, especially not RTL/LTR languages. You switch once or twice ,and you expect the reload because (even mentally, as a user) your page may (and does) end up looking radically different. I don't see a reason why it's crucial to do dynamic reloading to regular users, but even worse, doing it for RTL may actually increase your implementation problems by hundreds of percentages.

I would really highly recommend against reloading RTL/LTR language without refresh, and I would highly recommend against reinventing the wheel. 

My recommendation is that you start by using existing working tools as much as you possibly can, and try to identify the hardships/challenges you encounter along the way; make overrides if you have to, identify what is missing. 

Fixing up the existing tools to overcome the challenges you say you've identified may well be **orders of magnitude easier** than doing your own RTL support from scratch from now on. By far.

Either way, using existing tools is by far a much "safer to fail" experiment than creating your own from scratch, so I'd at least start with that and iterate forward.

santhoshtr:
@edg2s @Krinkle, There seems to be a misunderstanding about CX. Happy to clarify if required. The vue usage in CX for a new feature called [section translation](https://www.mediawiki.org/wiki/Content_translation/Section_translation). It is targetted at mobile users primarily in selected smaller wikis in certain geographies. Definitely not a replacement for whole CX or not going to put in front of all wiki users. The application will behave like an SPA, within a SpecialPage in those cases. It is the most active project by language team for ongoing year and the team is accountable for its development and deployment. Language team regularly publish the [status updates](https://www.mediawiki.org/wiki/Content_translation/Section_translation) to the organization and public too. When you say it is not deployable, I don't understand. A current version is available at https://sx.wmflabs.org/index.php/Special:ContentTranslation(needs login, work in progress, better in mobiles and smaller screens etc.). For more discussion, this pull request may not be the right place.

@Krinkle About the logical properties and browser support, I was not at all asking to start using logical properties right now and exclude some section of users. I was only referring how the general script directionality support in web is progressing. I was suggesting that we need to architecture this UI library in a way we use logical direction as design and component vocabulary. It is not about flipping left and right or using CSSJanus or not. It is about how we can be inclusive about these scripts from the design stage, to the developer abstractions and helpers provided in a UI library. Script directionality and support is more than CSS flipping. But unfortunately, we are just stuck with the implementation detail of RTL flipping so I don't know how we will achieve such goals. I have bigger dreams about supporting languages with different scripts better with wikipedia.

About the usecase of whether users will switch languages or not, at least for projects like Content translation where you work with at least two languages all the time, it is relevant. 

Talking about just the RTL flipping alone, ultimately we need RTL and LTR styles either bundled together(compresses well, can be generated at build step in UI library, UI library development+testing+demo+documentation can be done without ResourceLoader wiring) or have separate CSS assets, loaded on demand with the help of ResourceLoader wiring in a mediawiki context(Should a frontend component library have such strong dependencies is an architecture question, has impact on development process). The third option that is what I recommend for long term is design language and helpers that use logical directions. The implementation can use the current browser support or use flipping using CSS processors in progressive enhancement manner. This is the approach that popular UI libraries in Vue ecosystem already follows. This is what UI library in Section translation project tries to do.

To me, the whole discussion and debates are because, we miss a vision and plan for WVUI - Are we going to develop this like an elaborate UI library with futuristic goals? Or just for the search project now? Or iteratively patched up library on need basis of team in charge of development? How does this fit in the fronted architecture goals? I asked around these recently but did not get promising answers.

niedzielski:
Thank you all for the discourse. The relevant tickets for choosing the long term approach are covered, so far as I know, by these two:

- [Add CSSJanus support](https://phabricator.wikimedia.org/T261323)
- [Support a responsive grid system](https://phabricator.wikimedia.org/T90687)

Everyone's thoughts would be extremely welcome in the resolution of said tickets. My only pressing individual concern is that the library makes sense within and without MediaWiki. On Web alone, we have some notable experts (I'm thinking of @Volker-E and @j4n-co) and surely with the level of interest from @edg2s, @santhoshtr, @mooeypoo, @catrope, and @Krinkle, this patch has already generated, we'll pick the best possible solution in conjunction with WMDE (/cc @manicki) and other stakeholders. Although I look forward to participating in those discussions myself as well, I most look forward to benefiting from the solution picked.

I think that will take time though and this ticket is not the venue for continued discussion. The above would greatly benefit from the details discussed here so far so I would appreciate it if the contributors here would capture their thoughts in either / both of the above tickets.

It sounds like we have some good options available to us no less then CSSJanus (prebuild or runtime), mixins, or grid framework. For the purposes of search, the solution proposed in this patch, which adds a single Less mixin and several usages, satisfies the use case we need to ship so far as I know without implying a long term decision, as everyone seems to agree the a better approach will quite soon be needed. Please speak up if this is not the case.

The short term focus is search. The long term collaboration discussions and contributions can happen in parallel. For example, we've had excellent contributions from Anne, Santhosh, and others in the form of an icon component currently in use by search and that I think will be wanted by many in the long term as well.

In this interim time, I ask that contributors not attempt to unintentionally force a long term decision. I will reiterate that it is my hope to unblock the search work and move forward with the current modest solution rather than to unintentionally set a precedent which to me seems a lot harder to change than the purely technical work required by any future refactor.

edg2s:
>  I will reiterate that it is my hope to unblock the search work

I'm still not clear what is blocked here, given that the search component will live inside MediaWiki.

edg2s:
> For the purposes of search, the solution proposed in this patch, which adds a single Less mixin and several usages, satisfies the use case we need to ship so far as I know without implying a long term decision, as everyone seems to agree the a better approach will quite soon be needed. Please speak up if this is not the case.

As Moriel said, discarding 10+ years of learning for a new solution without a strong reason is going to lead to problems. While it may appear to work in a clean sandbox, the real world is messy, and we have tons of old code, hacks and user gadgets that we need to be compatible with.

Off the top of my head, adding the [dir=RTL] wrapper will change the specificity of every selector. Almost every time we touch the specificity of an OOUI component we cause a regression in a downstream override.

It may also not be compatible with directionality that is set via CSS property.

j4n-co:
To clarify the use-case of client-side only RTL flipping, we anticipate that the search component will eventually include some sort of language switcher. That component will enable searching across different languages from this one search box. In that scenario, switching between RTL and LTR on the client is a valid use-case. Additionally, a server-side refresh would not be desirable in that situation. To facilitate that behaviour, we either have to ship [dir=RTL] prefixed CSS, take advantage of direction/logical CSS like flex-box, but probably both.

To see how this scenario functions without either of these approaches, you can try searching in an RTL language on  https://www.wikipedia.org/ and notice the directionality of the results in incorrect. Since that's basically the same use-case 
that's in question here (search), I'd be curious as to how CSSJanus could be implemented to address this scenario. Would we have to ship two entirely separate RTL & LTR stylesheets down the wire?

edg2s:
> To clarify the use-case of client-side only RTL flipping, we anticipate that the search component will eventually include some sort of language switcher. That component will enable searching across different languages from this one search box. In that scenario, switching between RTL and LTR on the client is a valid use-case. Additionally, a server-side refresh would not be desirable in that situation. To facilitate that behaviour, we either have to ship [dir=RTL] prefixed CSS, take advantage of direction/logical CSS like flex-box, but probably both.
> 
> To see how this scenario functions without either of these approaches, you can try searching in an RTL language on https://www.wikipedia.org/ and notice the directionality of the results in incorrect. Since that's basically the same use-case
> that's in question here (search), I'd be curious as to how CSSJanus could be implemented to address this scenario. Would we have to ship two entirely separate RTL & LTR stylesheets down the wire?

That's definitely an interesting edge case. I haven't seen any designs on the tickets so I will you assume you are aiming for functionality similar to what is already on www.wiki.

The user is landing on an English LTR page and is expected to choose their language from the dropdown (by identifying "EN") then continue using the form. If you type in RTL the text input automatically flips, but the results are not:
image

Flipping the results is something you could code for manually, similar to the approach used here. I'm not sure flipping the rest of the widget would be warranted given (1) the user has already started interacting with it in LTR form and (2) flipping the widget underneath the user and moving the button they just clicked could be a clunky interaction. If product decides the rest of the widget needs to be flipped when you select an RTL language, that could be done by shipping both stylesheets as you say and swapping them out (toggling `[disabled]` attribute).

Aside from the maybe the multi-directional results, the rest sounds like an edge case for when you have a multi-lingual page. Accordingly the feature to flip individual widgets on the page should be implemented downstream for the foreseeable future; adding such an feature to this library would be the sort of irreversible decision that @niedzielski is warning against, as we'd be stuck with a [dir=rtl]-wrapper-type approach.

Whatever level of bidi selectors you end up with here, you should mark them with `/* @noflip */` so that the code is compatible with Janus, as depending how the library is bundled it may be difficult to disable Janus for a single file in MW.

niedzielski:
> I'm still not clear what is blocked here, given that the search component will live inside MediaWiki.

@edg2s, we're trying to build the search widget within this library. This patch is needed to ship. This discussion is blocking merging.

> As Moriel said, discarding 10+ years of learning for a new solution without a strong reason is going to lead to problems. While it may appear to work in a clean sandbox, the real world is messy, and we have tons of old code, hacks and user gadgets that we need to be compatible with.

It sounds like the worst case scenario is we'd fallback to turning on CSSJanus.

> Accordingly the feature to flip individual widgets on the page should be implemented downstream for the foreseeable future; adding such an feature to this library would be the sort of irreversible decision that @niedzielski is warning against, as we'd be stuck with a [dir=rtl]-wrapper-type approach.

All parties agree that a long term solution is wanted and the mixin is a stopgap that is inadequate in itself and doesn't risk becoming a long term decision the way a grid framework or CSSJanus might.

edg2s:
> All parties agree that a long term solution is wanted and the mixin is a stopgap that is inadequate in itself and doesn't risk becoming a long term decision the way a grid framework or CSSJanus might.

It does risk becoming a long term solution, because it fundamentally changes the way we detect if a widget is in an RTL context (by looking for attributes on any ancestor). Switching back to Janus would be difficult if this became expected behaviour.

> This patch is needed to ship.

There appears to be no good reason to prioritise RTL support outside of MediaWiki, especially as doing so is going to rush us into a contentious decision.

edg2s:
> by looking for attributes on any ancestor

Another places this mixin approach falls down is when you have a content direction that doesn't match the interface direction. This could be someone browsing an RTL wiki with an LTR interface language, or vice-versa (e.g. someone in mediawiki.org or commons using an Arabic interface). We also sometimes need to force components to be LTR, for example code input.

In various places we sometimes have to have interface elements that are attached to a content area. We currently handle this fine by only shipping the stylesheet matching the direction of the interface language, and ensuring that all interface areas are marked with the correct directionality.

In this new approach, the `[dir=rtl]` selector would always take precedence, so it would be impossible to correct the directionality of LTR components within RTL content. Adding `[dir=ltr]` would still not work, as the last selector written would take precedence, to the closest ancestor.

PR #77, opened 2020-08-18
[docs] Revise integrated development tips

niedzielski:
Revise tips for integrated development workflows, add some Input JSDocs,
and remove a test-only function.

### DIFF
--- a/src/components/input/Input.vue
+++ b/src/components/input/Input.vue
@@ -78,6 +78,10 @@ export const mwIconInfo = 'M9.5 16A6.61 6.61 0 0 1 3 9.5 6.61 6.61 0 0 1 9.5 3 6
 
 export default Vue.extend( {
 	name: 'WvuiInput',
+	/**
+	 * All atributes set on the components such as disabled and type are passed to the underlying
+	 * input.
+	 */
 	inheritAttrs: false,
 	props: {
 		value: {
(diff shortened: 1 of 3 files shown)


src/components/input/Input.vue
```diff
@@ -78,6 +78,10 @@ export const mwIconInfo = 'M9.5 16A6.61 6.61 0 0 1 3 9.5 6.61 6.61 0 0 1 9.5 3 6
 
 export default Vue.extend( {
 	name: 'WvuiInput',
+	/**
+	 * All atributes set on the components such as disabled and type are passed to the underlying
```
berndsi, line 82: typo.

BTW, I noticed another one below: `properyy` on line 110. Not sure if it's worth its own PR.

src/components/input/Input.vue
```diff
@@ -123,7 +128,7 @@ export default Vue.extend( {
 				'wvui-input--clearable': this.clearable
 			};
 		},
-		iconName(): string {
+		iconName(): string|undefined {
```
berndsi, line 131: Should there be a unit test associated with this return type fix?

niedzielski: It looks like this method was only used by tests. I've opted to delete the method instead.

berndsi: Even better. :)

berndsi:
@niedzielski Did you want to update the PR message before I squash and merge it? The "fix a return type" turned into something else.

speedandfunction-anribolon (approved):

niedzielski:
> The "fix a return type" turned into something else.

Revised :+1:

PR #79, opened 2020-08-25
[build] Split icons into a distinct entry

niedzielski:
- Stop exporting all icons from the WVUI entry. This is big and many
  consumers won't need it. Instead, output a distinct icons-only entry
  that is otherwise independent of the library.

- Clean the dist directory before building. The new build process
  contains two Webpack configurations: a general WVUI product and an
  icons-only product. However, the Webpack clean plugin does not appear
  to have great support for multi-configuration projects so the clean
  step is moved before building to avoid one config clobbering the
  outputs of the other.

- Divide Webpack configuration into two parts: common and build. The
  former is used by both Storybook and building the development /
  production outputs. The latter is used only to generate the build
  products. This change was driven by how clean is executed now: at
  import time (that is, with side-effects).

- Revise the bundlesize config.


.webpack/config.js
```diff
@@ -0,0 +1,139 @@
...
+
+/**
+ * @param {Parameters<webpack.ConfigurationFactory>[0]} _env
+ * @param {Parameters<webpack.ConfigurationFactory>[1]} argv
+ * @return {ReturnType<webpack.ConfigurationFactory>[]}
+ */
+module.exports = ( _env, argv ) => [
+	config( argv, 'wvui', { [ Chunk.Wvui ]: path.resolve( __dirname, '../src/entries/wvui.ts' ) } ),
+	config( argv, 'wvui-icons', { [ Chunk.WvuiIcons ]: path.resolve( __dirname, '../src/entries/wvui-icons.ts' ) } )
+];
```
speedandfunction-anribolon, line 139: I really like it @niedzielski! Outstanding!

speedandfunction-anribolon (approved):

PR #82, opened 2020-09-01
Combine fetch() and json()

berndsi:
Combine fetch() and json() so it can be easily replaced by another implementation, like jQuery.ajax().
Add optional parameter to the SearchClient so a client can pass in an alternative getJson() implementation.
If none is provided then native fetch is used.

### DIFF
--- a/src/components/typeahead-search/http/SearchClient.ts
+++ b/src/components/typeahead-search/http/SearchClient.ts
@@ -1,13 +1,17 @@
+import { GetJson } from '../../../http/fetch';
+
 export interface SearchClient {
 	/**
 	 * @param query search string to search for
 	 * @param domain the base URL for the wiki without protocol. Example: 'sr.wikipedia.org'
 	 * @param limit maximum number of results
+	 * @param fetchJsonImpl a function which returns a promise that resolves to a json
 	 */
 	fetchByTitle(
 		query: string,
 		domain: string,
-		limit?: number
+		limit?: number,
+		fetchJsonImpl?: GetJson
 	): Promise<SearchResponse>;
 }
 
--- a/src/components/typeahead-search/http/SearchClient.ts
+++ b/src/components/typeahead-search/http/SearchClient.ts
@@ -1,17 +1,13 @@
-import { GetJson } from '../../../http/fetch';
-
 export interface SearchClient {
 	/**
 	 * @param query search string to search for
 	 * @param domain the base URL for the wiki without protocol. Example: 'sr.wikipedia.org'
 	 * @param limit maximum number of results
-	 * @param fetchJsonImpl a function which returns a promise that resolves to a json
 	 */
 	fetchByTitle(
 		query: string,
 		domain: string,
-		limit?: number,
-		fetchJsonImpl?: GetJson
+		limit?: number
 	): Promise<SearchResponse>;
 }
 
--- a/src/components/typeahead-search/http/restSearchClient.ts
+++ b/src/components/typeahead-search/http/restSearchClient.ts
@@ -22,12 +22,12 @@ interface RestThumbnail {
 }
 
 function isRestResponse( response: unknown ): response is RestResponse {
-	return ( response as RestResponse ).pages !== undefined;
+	return Array.isArray( ( response as RestResponse ).pages );
 }
 
 function checkResponse( response: unknown ): Promise<RestResponse> {
 	if ( !isRestResponse( response ) ) {
-		return Promise.reject( 'Not a RestResponse' );
+		return Promise.reject( 'Not a valid RestResponse' );
 	}
 
 	return Promise.resolve( response as RestResponse );
--- a/src/components/typeahead-search/http/restSearchClient.ts
+++ b/src/components/typeahead-search/http/restSearchClient.ts
@@ -48,35 +48,26 @@ function adaptApiResponse( query: string, response: Record<string, unknown> ): S
 	};
 }
 
-// https://www.mediawiki.org/wiki/API:REST_API/Reference#Autocomplete_page_title
-function fetchByTitle(
-	query: string,
-	domain: string,
-	limit: number,
-	getJsonImpl: GetJson | undefined
-): Promise<SearchResponse> {
-	query = query.trim();
-	if ( !query ) {
-		return Promise.resolve( adaptApiResponse( query, { pages: [] } ) );
-	}
-	const params = {
-		q: query,
-		limit: limit
-	};
-	const headers = {
-		accept: 'application/json'
-	};
-
-	const url = `//${domain}/w/rest.php/v1/search/title?${buildQueryString( params )}`;
-
-	const getJson = !getJsonImpl ? fetchJson : getJsonImpl;
-	return getJson( url, { headers } ).then( ( json ) => adaptApiResponse( query, json ) );
-}
-
-export function restSearchClient(): SearchClient {
+export function restSearchClient( fetchJsonImpl?: GetJson ): SearchClient {
 	return {
-		fetchByTitle( query, domain, limit = 10, getJsonImpl = undefined ) {
-			return fetchByTitle( query, domain, limit, getJsonImpl );
+		// https://www.mediawiki.org/wiki/API:REST_API/Reference#Autocomplete_page_title
+		fetchByTitle( query, domain, limit = 10 ): Promise<SearchResponse> {
+			query = query.trim();
+			if ( !query ) {
+				return Promise.resolve( adaptApiResponse( query, { pages: [] } ) );
+			}
+			const params = {
+				q: query,
+				limit: limit
+			};
+			const headers = {
+				accept: 'application/json'
+			};
+
+			const url = `//${domain}/w/rest.php/v1/search/title?${buildQueryString( params )}`;
+
+			const getJson = !fetchJsonImpl ? fetchJson : fetchJsonImpl;
+			return getJson( url, { headers } ).then( ( json ) => adaptApiResponse( query, json ) );
 		}
 	};
 }
--- a/src/components/typeahead-search/http/restSearchClient.ts
+++ b/src/components/typeahead-search/http/restSearchClient.ts
@@ -57,9 +57,11 @@ export function restSearchClient( getJson: FetchJson = fetchJson ): SearchClient
 		// https://www.mediawiki.org/wiki/API:REST_API/Reference#Autocomplete_page_title
 		fetchByTitle( query, domain, limit = 10 ): Promise<SearchResponse> {
 			query = query.trim();
+
 			if ( !query ) {
 				return Promise.resolve( adaptApiResponse( query, { pages: [] } ) );
 			}
+
 			const params = {
 				q: query,
 				limit: limit
--- a/src/components/typeahead-search/http/restSearchClient.ts
+++ b/src/components/typeahead-search/http/restSearchClient.ts
@@ -1,4 +1,4 @@
-import { buildQueryString, fetchJson, GetJson } from '../../../http/fetch';
+import { buildQueryString, fetchJson, FetchJson } from '../../../http/fetch';
 import { SearchClient, SearchResponse } from './SearchClient';
 
 // https://www.mediawiki.org/wiki/API:REST_API/Reference#Search_result_object
@@ -40,7 +40,7 @@ function adaptApiResponse( query: string, restResponse: RestResponse ): SearchRe
 	};
 }
 
-export function restSearchClient( getJson: GetJson = fetchJson ): SearchClient {
+export function restSearchClient( getJson: FetchJson = fetchJson ): SearchClient {
 	return {
 		// https://www.mediawiki.org/wiki/API:REST_API/Reference#Autocomplete_page_title
 		fetchByTitle( query, domain, limit = 10 ): Promise<SearchResponse> {
--- a/src/components/typeahead-search/http/restSearchClient.ts
+++ b/src/components/typeahead-search/http/restSearchClient.ts
@@ -21,8 +21,7 @@ interface RestThumbnail {
 	height?: number | null;
 }
 
-function adaptApiResponse( query: string, response: Record<string, unknown> ): SearchResponse {
-	const restResponse: RestResponse = response as unknown as RestResponse;
+function adaptApiResponse( query: string, restResponse: RestResponse ): SearchResponse {
 	return {
 		query,
 		results:
@@ -59,7 +58,8 @@ export function restSearchClient( getJson: GetJson = fetchJson ): SearchClient {
 
 			const url = `//${domain}/w/rest.php/v1/search/title?${buildQueryString( params )}`;
 
-			return getJson( url, { headers } ).then( ( json ) => adaptApiResponse( query, json ) );
+			return getJson( url, { headers } )
+				.then( ( json ) => adaptApiResponse( query, json as RestResponse ) );
 		}
 	};
 }
--- a/src/components/typeahead-search/http/restSearchClient.ts
+++ b/src/components/typeahead-search/http/restSearchClient.ts
@@ -48,7 +48,7 @@ function adaptApiResponse( query: string, response: Record<string, unknown> ): S
 	};
 }
 
-export function restSearchClient( fetchJsonImpl?: GetJson ): SearchClient {
+export function restSearchClient( getJson: GetJson = fetchJson ): SearchClient {
 	return {
 		// https://www.mediawiki.org/wiki/API:REST_API/Reference#Autocomplete_page_title
 		fetchByTitle( query, domain, limit = 10 ): Promise<SearchResponse> {
@@ -66,7 +66,6 @@ export function restSearchClient( fetchJsonImpl?: GetJson ): SearchClient {
 
 			const url = `//${domain}/w/rest.php/v1/search/title?${buildQueryString( params )}`;
 
-			const getJson = !fetchJsonImpl ? fetchJson : fetchJsonImpl;
 			return getJson( url, { headers } ).then( ( json ) => adaptApiResponse( query, json ) );
 		}
 	};
--- a/src/components/typeahead-search/http/restSearchClient.ts
+++ b/src/components/typeahead-search/http/restSearchClient.ts
@@ -64,7 +64,7 @@ export function restSearchClient( getJson: FetchJson = fetchJson ): SearchClient
 
 			const params = {
 				q: query,
-				limit: limit
+				limit
 			};
 			const headers = {
 				accept: 'application/json'
--- a/src/components/typeahead-search/http/restSearchClient.ts
+++ b/src/components/typeahead-search/http/restSearchClient.ts
@@ -21,20 +21,6 @@ interface RestThumbnail {
 	height?: number | null;
 }
 
-function isRestResponse( response: unknown ): response is RestResponse {
-	return response instanceof Object &&
-		'pages' in response &&
-		Array.isArray( ( response as RestResponse ).pages );
-}
-
-function checkResponse( response: unknown ): Promise<RestResponse> {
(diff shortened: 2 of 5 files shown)


src/components/typeahead-search/http/SearchClient.ts
```diff
@@ -1,13 +1,17 @@
...
 	 * @param domain the base URL for the wiki without protocol. Example: 'sr.wikipedia.org'
 	 * @param limit maximum number of results
+	 * @param fetchJsonImpl a function which returns a promise that resolves to a json
 	 */
 	fetchByTitle(
 		query: string,
 		domain: string,
-		limit?: number
+		limit?: number,
+		fetchJsonImpl?: GetJson
```
niedzielski, line 14: Any reason not to request the fetch implementation in the search implementation constructor instead?

src/components/typeahead-search/http/SearchClient.ts
```diff
@@ -1,13 +1,17 @@
+import { GetJson } from '../../../http/fetch';
+
 export interface SearchClient {
 	/**
 	 * @param query search string to search for
 	 * @param domain the base URL for the wiki without protocol. Example: 'sr.wikipedia.org'
 	 * @param limit maximum number of results
+	 * @param fetchJsonImpl a function which returns a promise that resolves to a json
```
niedzielski, line 8: Nitpick: casing on "json" should be upper.

src/components/typeahead-search/http/restSearchClient.ts
```diff
@@ -59,15 +68,15 @@ function fetchByTitle(
...
+
+	const getJson = !getJsonImpl ? fetchJson : getJsonImpl;
+	return getJson( url, { headers } ).then( ( json ) => adaptApiResponse( query, json ) );
 }
 
 export function restSearchClient(): SearchClient {
 	return {
-		fetchByTitle( query, domain, limit = 10 ) {
-			return fetchByTitle( query, domain, limit );
+		fetchByTitle( query, domain, limit = 10, getJsonImpl = undefined ) {
```
niedzielski, line 78: Here we default to undefined but above we default to fetchJson. Is it possible to stick to one or the other?

src/http/fetch.ts
```diff
@@ -1,9 +1,26 @@
...
 // Currently this rejects the returned promise if window.fetch is not available in the browser.
-// The plan is to add a fallback so we can support older browsers in the future.
-export function fetch( resource: string, init?: RequestInit ): Promise<Response> {
+// The plan is to add a way for clients of this library to pass in an implementation, e.g.
+// using jQuery.ajax(), so we can support older browsers in the future.
+export function fetchJson(
+	resource: string,
+	init?: RequestInit
+): Promise<Record<string, unknown>> {
 	if ( 'fetch' in window ) {
```
niedzielski, line 15: Do you think this check still makes sense? If the client doesn't specify a Fetch implementation they're implicitly saying that window.fetch is available.

src/http/fetch.ts
```diff
@@ -1,9 +1,26 @@
...
-		return window.fetch( resource, init );
+		return window.fetch( resource, init )
+			.then( ( response ) => {
+				if ( response.ok ) {
+					return response.json();
+				} else {
+					return {};
```
niedzielski, line 21: Should we reject here? If not, please consider adding a comment inline.

src/http/fetch.ts
```diff
@@ -1,9 +1,26 @@
-// A wrapper for native fetch() in browsers.
+// a function which returns a promise that resolves to a json
+export type GetJson = (
```
niedzielski, line 2: Would "FetchJson"  naming improve the parity with the fetchJson implementation below?

src/http/fetch.ts
```diff
@@ -1,9 +1,26 @@
...
+	resource: string,
+	init?: RequestInit
+) => Promise<Record<string, unknown>>;
+
+// A wrapper which combines native fetch() in browsers and the following json() call.
 // Currently this rejects the returned promise if window.fetch is not available in the browser.
-// The plan is to add a fallback so we can support older browsers in the future.
-export function fetch( resource: string, init?: RequestInit ): Promise<Response> {
+// The plan is to add a way for clients of this library to pass in an implementation, e.g.
```
niedzielski, line 9: We can probably describe the intended operation as it is now since we're building what we need for today.

src/http/fetch.ts
```diff
@@ -1,9 +1,26 @@
-// A wrapper for native fetch() in browsers.
+// a function which returns a promise that resolves to a json
+export type GetJson = (
+	resource: string,
+	init?: RequestInit
+) => Promise<Record<string, unknown>>;
```
niedzielski, line 5: Does this always return a record? I would guess that we can only assume a JSON type here.

src/components/typeahead-search/http/restSearchClient.ts
```diff
@@ -21,7 +21,15 @@ interface RestThumbnail {
...
-function adaptApiResponse( query: string, restResponse: RestResponse ): SearchResponse {
+function adaptApiResponse( query: string, response: Record<string, unknown> ): SearchResponse {
+	if ( !( 'pages' in response ) ) {
+		return {
+			query,
+			results: []
+		};
+	}
+
+	const restResponse: RestResponse = response as unknown as RestResponse;
```
niedzielski, line 32: Should we reject the outer Promise if a bad type is passed in?

src/http/fetch.ts
berndsi: @niedzielski I really want a JSON object here, so I tried `object`, which lint is unhappy with:
> Don't use `object` as a type. The `object` type is currently hard to use ([see this issue](https://github.com/microsoft/TypeScript/issues/21732)).
Consider using `Record<string, unknown>` instead, as it allows you to more easily inspect and use the keys  @typescript-eslint/ban-types

Trying `Object` lead me to this:
> Don't use `Object` as a type. The `Object` type actually means "any non-nullish value", so it is marginally better than `unknown`.
- If you want a type meaning "any object", you probably want `Record<string, unknown>` instead.
- If you want a type meaning "any value", you probably want `unknown` instead  @typescript-eslint/ban-types

niedzielski: It's my understanding that TypeScript's `object` won't cover all valid JSON types like `null`, just like `Record`. If the response might be `null` or any non-object type, neither `Record` nor `object` seems like a good choice. Maybe `unknown` is the best we can do for now.

https://www.typescriptlang.org/docs/handbook/basic-types.html#object
https://www.json.org
https://github.com/microsoft/TypeScript/issues/1897

src/components/typeahead-search/http/SearchClient.ts
berndsi: Good suggestion. I've moved it to become a parameter of the search implementation constructor.

src/components/typeahead-search/http/SearchClient.ts
berndsi: That code got removed. I've updated the other occurrence of this.

src/components/typeahead-search/http/restSearchClient.ts
berndsi: Excellent idea. That simplifies some of the code.

src/http/fetch.ts
berndsi: I've taken it out.

src/http/fetch.ts
berndsi: Sounds good.

src/http/fetch.ts
berndsi: Done.

src/http/fetch.ts
berndsi: FetchJson sounds good to me. The drawback would be that it's the same as fetchJson if you ignore the former starting with an uppercase letter. What do you think about renaming `fetchJson` to `nativeFetchJson`?

src/components/typeahead-search/http/restSearchClient.ts
berndsi: Do you mean with the `if ( !( 'pages' in response ) )` I had above?

src/http/fetch.ts
```diff
@@ -1,12 +1,24 @@
...
+// A wrapper which combines native fetch() in browsers and the following json() call.
+export function fetchJson(
+	resource: string,
+	init?: RequestInit
+): Promise<Record<string, unknown>> {
+	return fetch( resource, init )
+		.then( ( response ) => {
+			if ( response.ok ) {
+				return response.json();
+			} else {
```
niedzielski, line 16: Very minor: the else here is optional given the return statement on the line above.

src/http/fetch.ts
niedzielski: Personally, I like `FetchJson` with the thinking that PascalCase signals types and cameCase signals instances. Not a big deal either way.

src/http/fetch.ts
```diff
@@ -1,12 +1,24 @@
...
-		return Promise.reject( 'window.fetch() not available' );
-	}
+// A function which returns a promise that resolves to a JSON object
+export type GetJson = (
+	resource: string,
+	init?: RequestInit
+) => Promise<Record<string, unknown>>;
+
+// A wrapper which combines native fetch() in browsers and the following json() call.
+export function fetchJson(
```
niedzielski, line 8: It occurs to me that we will probably want the ability to cancel requests. How do you think we should represent that? In Popups, we [added a method to the Promise object](https://gerrit.wikimedia.org/g/mediawiki/extensions/Popups/+/master/src/gateway/rest.js#77) but I'm not sure if that's a good idea or not here.

src/components/typeahead-search/http/restSearchClient.ts
niedzielski: What I mean is that it felt a little strange to say that we expect a generic parameter, and then we test that parameter for the pages property, and then we still cast it here. Personally, I normally use [type guards](https://www.typescriptlang.org/docs/handbook/advanced-types.html#using-type-predicates) which are usually imperfect but at least capture the idea without a cast. There's at least one example in [WVUI's InputType](https://github.com/wikimedia/wvui/blob/master/src/components/input/InputType.ts#L13).

src/http/fetch.ts
berndsi: @niedzielski Ok, switching from Record<string, unknown> to unknown looks simpler.

src/http/fetch.ts
berndsi: Done

src/http/fetch.ts
berndsi: Done

src/http/fetch.ts
berndsi: This requires some more thought. Can this be done in a separate PR?

src/http/fetch.ts
berndsi: Replaced it with a guard clause.

src/http/fetch.ts
speedandfunction-anribolon: Just for the reference, it will require adding 2 different implementations. $.getJson() returns a promise with jqXHR object, which has `abort()` method (as @niedzielski mentioned above). `fetch` implementation will require using [AbortController](https://developer.mozilla.org/ru/docs/Web/API/AbortController). I think it's a good idea to split it to a separate patch @niedzielski @berndsi

src/components/typeahead-search/http/restSearchClient.ts
```diff
@@ -40,34 +40,26 @@ function adaptApiResponse( query: string, restResponse: RestResponse ): SearchRe
...
-	};
-	const headers = {
-		accept: 'application/json'
-	};
+export function restSearchClient( getJson: FetchJson = fetchJson ): SearchClient {
+	return {
+		// https://www.mediawiki.org/wiki/API:REST_API/Reference#Autocomplete_page_title
+		fetchByTitle( query, domain, limit = 10 ): Promise<SearchResponse> {
+			query = query.trim();
+			if ( !query ) {
```
speedandfunction-anribolon, line 48: Minor: Consider adding extra line before `if` statement, It's a good practice.

speedandfunction-anribolon (commented):
LGTM! Just a small comment inline.

src/components/typeahead-search/http/restSearchClient.ts
berndsi: @niedzielski I've pushed my first attempt. It currently still feels a bit weird since it also involves a cast:
```
function isRestResponse( response: unknown ): response is RestResponse {
	return ( response as RestResponse ).pages !== undefined;
}
```

I've tried something like this:
```
function isRestResponse( response: unknown ): response is RestResponse {
	return 'pages' in response;
}
```
but that results in `TS2571: Object is of type 'unknown'`.

src/components/typeahead-search/http/restSearchClient.ts
berndsi: I've added one before and after. IMO the one afterwards is more important since the if block includes a return statement.

src/http/fetch.ts
berndsi: Yes, the other reason is that we need to look at the browser compatibility table of [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController#Browser_compatibility). AbortController is marked as experimental.

niedzielski: Let's explore in a new PR :+1: Maybe we can abort when available (either AbortController or XHR / jQuery-based) and reject the promise otherwise.

src/components/typeahead-search/http/restSearchClient.ts
```diff
@@ -21,6 +21,18 @@ interface RestThumbnail {
 	height?: number | null;
 }
 
+function isRestResponse( response: unknown ): response is RestResponse {
```
niedzielski, line 24: I think you can avoid the case with something like this:

```ts
function isRestResponse( response: unknown ): response is RestResponse {
  return response instanceof Object && 'pages' in response && Array.isArray( response['pages'] )
}
```

src/components/typeahead-search/http/restSearchClient.ts
```diff
@@ -21,6 +21,18 @@ interface RestThumbnail {
...
+	return Array.isArray( ( response as RestResponse ).pages );
+}
+
+function checkResponse( response: unknown ): Promise<RestResponse> {
+	if ( !isRestResponse( response ) ) {
+		return Promise.reject( 'Not a valid RestResponse' );
+	}
+
+	return Promise.resolve( response as RestResponse );
```
niedzielski, line 33: Do we need this case here? The type guard is supposed to infer the type.

src/components/typeahead-search/http/restSearchClient.ts
```diff
@@ -21,6 +21,18 @@ interface RestThumbnail {
 	height?: number | null;
 }
 
+function isRestResponse( response: unknown ): response is RestResponse {
+	return Array.isArray( ( response as RestResponse ).pages );
+}
+
+function checkResponse( response: unknown ): Promise<RestResponse> {
+	if ( !isRestResponse( response ) ) {
+		return Promise.reject( 'Not a valid RestResponse' );
```
niedzielski, line 30: Extremely minor: consider rewording to the terser "Invalid RestResponse."

src/components/typeahead-search/http/restSearchClient.ts
```diff
@@ -40,34 +52,29 @@ function adaptApiResponse( query: string, restResponse: RestResponse ): SearchRe
...
+				return Promise.resolve( adaptApiResponse( query, { pages: [] } ) );
+			}
 
-export function restSearchClient(): SearchClient {
-	return {
-		fetchByTitle( query, domain, limit = 10 ) {
-			return fetchByTitle( query, domain, limit );
+			const params = {
+				q: query,
+				limit: limit
```
niedzielski, line 67: Consider shorthand for limit: `{q: query, limit }`.

niedzielski (changes requested):
Looking good. Just a couple more notes inline but you're close to the finish line from my perspective :+1:

src/components/typeahead-search/http/restSearchClient.ts
berndsi: Our eslint rules don't like `Array.isArray( response[ 'pages' ] )`:
```
 ["pages"] is better written in dot notation  dot-notation
```

niedzielski: Maybe use a line disable? Generally, that lint is a valid style criticism but TypeScript only recognizes the indexing syntax for this scenario.

src/components/typeahead-search/http/restSearchClient.ts
berndsi: Good call. Removed the cast.

src/components/typeahead-search/http/restSearchClient.ts
berndsi: Done

src/components/typeahead-search/http/restSearchClient.ts
berndsi: Done. Good catch!

src/components/typeahead-search/http/restSearchClient.ts
speedandfunction-anribolon: Please correct me if I'm wrong, but I think we should stand at the point where we know what exactly rest api returns. This is not that kind of rest Api Client that is considered to be a part of the library.

niedzielski: I suppose that's fair. An API is a contract so hopefully the MediaWiki REST folks will adhere to it. Soo... Cast I guess?

src/components/typeahead-search/http/restSearchClient.ts
```diff
@@ -40,34 +54,29 @@ function adaptApiResponse( query: string, restResponse: RestResponse ): SearchRe
 	};
```
speedandfunction-anribolon, line 54: I think it might be simplified a bit:
```
restResponse.pages
				.map( ( {thumbnail, ...theRestOfPage} ) => ( {
					...theRestOfPage
					thumbnail: thumbnail ? {
						url: thumbnail.url,
						width: thumbnail.width ?? undefined,
						height: thumbnail.height ?? undefined
					} : undefined
				} ) )
```

speedandfunction-anribolon: Or just for to be more precise
restResponse.pages
				.map( ( {thumbnail, id, key, title, description} ) => ( {
					id,
                                        key,
                                        title,
                                        description,
					thumbnail: thumbnail ? {
						url: thumbnail.url,
						width: thumbnail.width ?? undefined,
						height: thumbnail.height ?? undefined
					} : undefined
				} ) )

berndsi: LGTM.

src/components/typeahead-search/http/restSearchClient.ts
berndsi: Are you saying I should remove some of these checks or leave as is?

niedzielski: I'm saying to drop this function and sorry for my poor direction :]

@speedandfunction-anribolon is right. If this function's purpose is to check that the API is valid, it's not a real API because they're not adhering to the contract. If that's the case, we can leave this code in but should file a bug report. If that's not the case, we should be able to just cast from unknown to RestResponse without any checks.

berndsi: @niedzielski @speedandfunction-anribolon Ok, I've removed this code now. It's back to just the cast, as it was before

src/components/typeahead-search/http/restSearchClient.ts
berndsi: I made that change yesterday.

PR #83, opened 2020-09-01
eslint: Remove ecmaVersion properites

edg2s:
These are already inherited from wikimedia/language/* configs.


niedzielski:
Thank you!

PR #86, opened 2020-09-03
Blubber Config + Docker setup

nikkhn:
This is an initial go at a blubber config and docker-compose setup. I am by no means an expert so feel free to poke/pry. I also removed a few things that I believe are taken care of by eslint-config-wikimedia npm module.

### DIFF
--- a/.pipeline/blubber.yaml
+++ b/.pipeline/blubber.yaml
@@ -1,17 +1,17 @@
 version: v4
 lives:
-  in: /srv/service
+    in: /srv/service
 variants:
-  build:
-    base: docker-registry.wikimedia.org/nodejs-devel
-    copies: [local]
-    apt:
-      packages: [ git, build-essential, python-pkgconfig ] # needed for npm install
-    node:
-      requirements: [ package.json, package-lock.json]
-      env: development # unnecessary? 
-    builder:
-      command: [ npm, build ]
-  test:
-    includes: [ build ]
-    entrypoint: [ npm, test ]
+    build:
+        base: docker-registry.wikimedia.org/nodejs-devel
+        copies: local
+        apt:
+            packages: [git, build-essential, python-pkgconfig] # needed for npm install
+            node:
+                requirements: [package.json, package-lock.json]
+                env: development
+            builder:
+                command: [npm, build]
+    test:
+        includes: build
+        entrypoint: [npm, test]
--- a/.pipeline/blubber.yaml
+++ b/.pipeline/blubber.yaml
@@ -1,0 +1,15 @@
+version: v4
+lives:
+  in: /srv/service
+variants:
+  dev:
+    base: docker-registry.wikimedia.org/nodejs-devel
+    # copies: [local]
+    apt:
+      packages: [ git, build-essential, python-pkgconfig ] # needed for npm install
+    node:
+      requirements: [package.json, package-lock.json]
+      env: development # maybe unnecessary 
+    builder:
+      command: [npm, build]
+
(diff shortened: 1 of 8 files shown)


niedzielski:
@nikkhn, I'm guessing it's ok to reopen this.

.stylelintignore
```diff
@@ -3,6 +3,8 @@
 /docs/
 
 *.json
+*.yaml
+Dockerfile
```
niedzielski, line 7: Would you mind using an anchored path for the Dockerfile? `/Dockerfile`. It's definitely not a big deal but the rationale is that we've been erring on side of precision in trying to keep our ignore files very intentional and specific.

Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM docker-registry.wikimedia.org/nodejs10-devel
+RUN apt-get update && \
+	apt-get install -y \
+		build-essential \
+		# ca-certificates \
+		python-pkgconfig \
+		git
+RUN npm install
```
niedzielski, line 8: Is this the recommended command from RelEng / WMDE or should we use `npm ci`?

.pipeline/blubber.yaml
```diff
@@ -0,0 +1,17 @@
+version: v4
```
niedzielski, line 1: Please add a new section to the readme for Docker. The section should cover at least what Docker is used for in this repo and how to maintain it. For example, if adding a dependency now requires rebuiding the Docker image each time, the readme should cover that. If Docker is optional for local development, which I believe it is, the section should explicit call that out.

nikkhn:
@niedzielski I actually think I will just cherry-pick my commits into the gerrit repo like you suggested today. (Was trying to be slightly lazy and just keep this branch open and merge it in github haha)

.pipeline/blubber.yaml
```diff
@@ -0,0 +1,17 @@
...
+    build:
+        base: docker-registry.wikimedia.org/nodejs-devel
+        copies: [local]
+        apt:
+            packages: [git, build-essential, python-pkgconfig] # needed for npm install
+        node:
+            requirements: [package.json, package-lock.json]
+            env: development
+        builder:
+            command: [npm, build]
```
niedzielski, line 14: Does this line run `npm build` or `npm run build`? I thought we wanted the latter but could be mistaken.

package.json
```diff
@@ -46,7 +46,7 @@
     "format:js": "npm -s run linter:js -- --fix .",
     "format:styles": "npm -s run linter:styles -- --fix .",
     "linter:etc": "prettier .",
-    "linter:js": "eslint --cache --max-warnings 0 --report-unused-disable-directives --ext .js,.json,.ts,.vue",
```
niedzielski, line 49: Can you tell me what this change is for?

package-lock.json
```diff
@@ -6260,6 +6260,16 @@
         "del": "^4.1.1"
```
niedzielski, line 6260: From what I see, there are no changes to dependencies in this PR so this file should not be changed.

docker-compose.yaml
```diff
@@ -0,0 +1,16 @@
...
+            dockerfile: ./Dockerfile
+        image: wikimedia/wvui
+        volumes:
+            - ./:/app
+        working_dir: /app
+    storybook:
+        extends:
+            service: node
+        ports:
+            - "3003:3003"
```
niedzielski, line 15: Is this to match Storybook? If so, please add a comment.

niedzielski (changes requested):

niedzielski:
@nikkhn, just saw your comment. Works for me :+1:

nikkhn:
@niedzielski I will copy over your pr comments to gerrit and address there so they are documented

